PR #16, opened 2019-06-19
Kill GC'ed processes

### LINKED ISSUE #8: Kill garbage-collected processes
That's it. All processes should be killed on gc. Garbage should be collected after every test group. And at exit too.

printercu:
Fixes #8

### DIFF
--- a/bin/luatest
+++ b/bin/luatest
@@ -1,4 +1,25 @@
 #!/usr/bin/env tarantool
 
-local runner = require('luatest').runner
-os.exit(runner:run())
+local fun = require('fun')
+
+-- Tarantool does not invoke Lua's GC-callbacks on exit.
+-- So this method clears every loaded package and invokes GC explicitly.
+-- It does not clear previously loaded packages to not break any GC-callback
+-- which relies on built-in package.
+local function gc_sandboxed(fn)
+    local original = fun.iter(package.loaded):map(function(x) return x, true end):tomap()
+    local result = fn()
+    -- Collect list of new packages to not modify map while iterating over it.
+    local new_packages = fun.iter(package.loaded):
+      map(function(x) return x end):
+      filter(function(x) return not original[x] end):
+      totable()
+    for _, name in pairs(new_packages) do
+      package.loaded[name] = nil
+    end
+    collectgarbage()
+    return result
+end
+
+local result = gc_sandboxed(function() return require('luatest').runner:run() end)
+os.exit(result)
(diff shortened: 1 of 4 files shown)


luatest/process.lua
```diff
@@ -17,9 +18,15 @@ local function to_const_char(input)
...
+--- Starts process and returns immediately, not waiting until process is finished.
+-- @param path Executable path.
+-- @param[opt] args
+-- @param[opt] env
+-- @param[opt] options
+-- @param[opt] options.chdir Directory to chdir into before starting process.
+-- @param[opt] options.ignore_gc Don't install handler which kills GC'ed processes.
 function Process:start(path, args, env, options)
-    checks('table', 'string', '?table', '?table', {chdir = '?string'})
+    checks('table', 'string', '?table', '?table', {chdir = '?string', ignore_gc = '?boolean'})
```
rosik: How can it be used?

bin/luatest
```diff
@@ -1,4 +1,25 @@
...
-local runner = require('luatest').runner
-os.exit(runner:run())
+local fun = require('fun')
+
+-- Tarantool does not invoke Lua's GC-callbacks on exit.
+-- So this method clears every loaded package and invokes GC explicitly.
+-- It does not clear previously loaded packages to not break any GC-callback
+-- which relies on built-in package.
+local function gc_sandboxed(fn)
+    local original = fun.iter(package.loaded):map(function(x) return x, true end):tomap()
```
rosik: It would be better if you cleaned loaded packages after every test file. So they don't interfere

luatest/process.lua
```diff
@@ -45,14 +52,26 @@ end
...
+    self.kill_pid(self.pid, signal)
+end
+
+function Process.kill_pid(pid, signal, options)
     -- Signal values are platform-dependent so we can not use ffi here
     signal = signal or 15
-    local exit_code = os.execute('kill -' .. signal .. ' ' .. self.pid)
-    if exit_code ~= 0 then
+    local exit_code = os.execute('kill -' .. signal .. ' ' .. pid .. ' 2> /dev/null')
+    if exit_code ~= 0 and (not options or not options.quiet) then
```
rosik: `and not (options and options.quiet)`

Also it lacks the type check.

luatest/process.lua
printercu: `{ignore_gc = true}` ?
This option:
- provides control over very specific feature,
- documents code, and make it easier to understand what's going on in  the code.

bin/luatest
printercu: 👎 for this. Also there can be common `test/helper.lua` file which is required before the whole suite.

PR #20, opened 2019-06-27
Use fio for non-blocking read in capture

printercu:
This eliminates complex logic with writing to pipe before reading. This also makes it possible to read from pipe which is already full, while writing to full pipe would block process completely.

Despite it uses private fio API, code gets much clearer.


printercu:
It works on macos :(

PR #21, opened 2019-06-27
Use real time instead of CPU time for duration

printercu:


Totktonada:
It is possible to track both, why not? BTW, I use clock.monotonic() for benchmarking purposes.

printercu:
@Totktonada we don't want to add much changes to luaunit for now. Real time is what user expects to see and it's very surprising to see time 0.01 while test was actually running for about 2 sec.

I think os.clock() was used because there is no `clock` module in vanilla lua/luajit and os.time() provides time without necessary precision. CPU time is not very useful in tests - jit most likely is not applied in most places and there is a lot of other code which can be not optimized for performance.

Totktonada:
I don't insist, just share my observations and personal experience.

PR #26, opened 2019-07-16
 Don't run suite hooks when suite is not going to be run

printercu:


luatest/capture.lua
```diff
@@ -130,27 +131,26 @@ function Capture:flush()
 end
 
 -- Run function with enabled/disabled capture and restore previous state.
--- In the case of error it prints error to original stdout.
+-- In the case of error it wraps  it into map-table with captured output added.
```
sharonovd: extra space

PR #40, opened 2019-10-02
Add custom table assertions

RunsFor:
This patch add several custom assertions which are intended to make table
checks more readable:

```
assert_table_has_key(tbl, key, extra_msg_or_nil)
assert_table_has_keys(tbl, ...)
assert_table_has_pair(tbl, key, value, extra_msg_or_nil)
assert_table_to_include(tbl, inclusion, extra_msg_or_nil)
```

```bash
$ bin/luatest -v asserts
Started on Wed Oct  2 21:54:53 2019
    asserts.test_assert_table_has_key ... Ok
    asserts.test_assert_table_has_keys ... Ok
    asserts.test_assert_table_has_pair ... Ok
    asserts.test_assert_table_to_include ... Ok
=========================================================
Ran 4 tests in 0.001 seconds, 4 successes, 0 failures
OK
```

### DIFF
--- a/luatest/luaunit.lua
+++ b/luatest/luaunit.lua
@@ -1761,6 +1761,51 @@ function M.assertNotIsMinusZero(value, extra_msg_or_nil)
 end
 
 ----------------------------------------------------------------
+--                     Custom assertions
+----------------------------------------------------------------
+
+function M.assert_table_has_key(tbl, key, extra_msg_or_nil)
+    M.assert_is_table(tbl)
+    M.assert_is_string(key)
+
+    if tbl[key] == nil then
+        fail_fmt(2, extra_msg_or_nil,
+            "expected %s to have key %s", prettystr(tbl), prettystr(key))
+    end
+end
+
+function M.assert_table_has_keys(tbl, ...)
+    M.assert_is_table(tbl)
+
+    local n = select('#', ...)
+    for i = 1,n do
+        local key = select(i, ...)
+        M.assert_table_has_key(tbl, key)
+    end
+end
+
+function M.assert_table_has_pair(tbl, key, value, extra_msg_or_nil)
+    M.assert_is_table(tbl)
+    M.assert_is_string(key)
+    M.assert_not_is_nil(value)
+
+    if tbl[key] ~= value then
+        fail_fmt(2, extra_msg_or_nil,
+            "expected %s to have key-value pair %s",
+            prettystr(tbl), prettystr({ [key] = value }))
+    end
+end
+
+function M.assert_table_to_include(tbl, inclusion, extra_msg_or_nil)
+    M.assert_is_table(tbl)
+    M.assert_is_table(inclusion)
+
+    for key, value in pairs(inclusion) do
+        M.assert_table_has_pair(tbl, key, value, extra_msg_or_nil)
+    end
+end
+
+----------------------------------------------------------------
 --                     Compatibility layer
 ----------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


printercu:
Have you seen assert_covers?

printercu:
Hi! Looks like `assert_table_has_pair` and `assert_table_to_include` are already represented by `assert_covers`, aren't they?

Where can this be useful to check that key exists without checking it's value? There can be corner cases, and I think user have to handle them himself. I see that proposed method uses `== nil`, so `assert_table_has_key({key = box.NULL}, 'key')` would mistakenly fail. WDYT about using `t.assert(tb[key], msg)` instead of this?

RunsFor:
Thanks for the review!

> Looks like assert_table_has_pair and assert_table_to_include are already represented by assert_covers, aren't they?

Thanks! Did not notice that. Maybe we just need to improve error messages for `assert_covers`

I decided to add those assertion helpers mostly because I need more readable error messages from luatest. Of course, user can do all the formatting, but such tests will loose in readability.

See the difference:

```lua
g.test_covers = function()
	t.assert_covers({foo = 'bar', bar = 'buzz'}, { buzz = 'foo' })
end

g.test_inclusion = function()
	t.assert_table_to_include({foo = 'bar', bar = 'buzz'}, { buzz = 'foo' })
end
```

Output:
```
Started on Thu Oct  3 22:19:37 2019
    asserts.test_covers ... FAIL
/Users/i.konyuhov/Workspace/luatest/luatest/init.lua:40: expected: {buzz="foo"}
actual: {}
    asserts.test_inclusion ... FAIL
/Users/i.konyuhov/Workspace/luatest/luatest/luaunit.lua:1804: expected {bar="buzz", foo="bar"} to have key-value pair {buzz="foo"}
```

How can I briefly understand what first test is actually testing without reading the test? Imaging if actual value is not hardcoded.

> assert_table_has_key({key = box.NULL}, 'key') would mistakenly fail

Why you think it should not fail?

> Where can this be useful to check that key exists without checking it's value?

I am partly agree with you. Maybe extended version would make more sense (with proper error formatting):

```lua
t.assert_table_has_key(tbl, 'foo', 'string')
-- or
t.assert_table_has_string_key(tbl, 'foo')
t.assert_table_has_string_keys(tbl, 'foo', 'bar', 'buzz')
-- etc
```

printercu:
> Why you think it should not fail?

Because the table actually has the key :) It'll be present in `pairs`, json string, saved to DB, etc.

> t.assert_table_has_key(tbl, 'foo', 'string')

Why do you need this new matchers while they can be expressed with `t.assert_covers(tbl, {foo = 'string'})`. Maybe we can just improve failure message for this matcher.

RunsFor:
> Why do you need this new matchers while they can be expressed with t.assert_covers(tbl, {foo = 'string'})?

`t.assert_covers` is useful when I want to check values. Besides values check sometimes I want to just check its type. Imaging `uuid` field was generated on the server side and I just need to verify, if it is a string.

printercu:
`t.assert_equals(type(tb.uuid), 'string')`?

RunsFor:
No,

Firstly, it would require more code for user to right. Secondly, it does not render readable unified error messages

printercu:
Please provide exact proposed matchers signatures. Maybe with some examples. Let's ask somebody else to take a look and give a feedback.

RunsFor:
I'm closing this PR.

Generally I wanted to have more matchers that both easy using, and generates readable error messages.

Currently I'm working on a simple module, which tries to mimic [rspec matchers](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers) which meets both criteria (imho).

printercu:
Great! I like the rspec way.

PR #47, opened 2019-10-21
 Make --shuffle option accept `group`, `all`, `none` values

### LINKED ISSUE #29: Test order is unexpected and misleading
Yes, sometimes it's useful. But expected behavior is test will be launched in user-defined order.

It’s especially important for tests in group.

printercu:
Fixes #29


rosik:
I would call it `--shuffle none | group | all`, and I think `--shuffle-seed` option should be separate.

PR #73, opened 2019-12-12
Prefix and colorize child processes' output

printercu:


knazarov:
I believe it would be a good idea to detect if the stdout is a tty, and only colorize output in that case.

printercu:
`tty` detection can be implemented later using `ffi.C.isatty`. Now it's not straightforward because colouring is used not only in OutputBeatifier but also in basic reports for errors and stats.

It should be considered to enable colouring by default in tty consoles and have flag to force colouring (e.g. for CI builds).

PS. colored log files can be viewed with `less -R`.

PR #75, opened 2019-12-16
Fix type check in `assert_covers` matcher

RunsFor:
Make `assert_covers` matcher to fail with appropriate message
instead of:

```
bad argument #1 to 'pairs' (table expected, got nil)
```


printercu:
Thanks! Can you please add same assertion for assert_not_covers as well?

RunsFor:
@printercu Done

PR #99, opened 2020-03-03
Add TARANTOOL_DIR to rockspec build.variables

HustonMmmavr:


printercu:
Note. TARANTOOL_DIR along with external_dependencies.TARANTOOL.header is required for building rock when tarantool is installed in custom directory.

PR #113, opened 2020-10-16
fix io to fio hack

### LINKED ISSUE #112: test: io_to_fio hack does not work anymore
https://github.com/tarantool/luatest/blob/ace5fbfa53af26f17c12254cb80ad5e42ddc752a/test/capture_test.lua#L21-L31

```
2: 1) capture.test_flush_large_strings
2: /build/usr/src/debug/luatest-0.5.2.2/test/capture_test.lua:29: bad argument #2 to 'setmetatable' (nil or table expected)
2: stack traceback:
2: 	/build/usr/src/debug/luatest-0.5.2.2/test/capture_test.lua:29: in function 'io_to_fio'
2: 	/build/usr/src/debug/luatest-0.5.2.2/test/capture_test.lua:37: in function 'capture.test_flush_large_strings'
2: 	...
2: 	[C]: in function 'xpcall'
```

The output comes from the CI [job](https://travis-ci.com/github/tarantool/luatest/jobs/400613565). tarantool-1.10 repository is enabled here, but 2.4 from EPEL is installed (see [how](https://github.com/tarantool/tarantool-php/blob/a631b56eba38ec3572c393b87ab54edf14b38aa2/test.sh#L38-L47) I overcome this kind of problems in CI).

It seems, the reason is https://github.com/tarantool/tarantool/commit/3d5b4daaa05c00152cb147c48c83c35ae1fd7981

So it should fail on the following tarantool versions:

* 2.4 since 2.4.0-133-g3d5b4daaa.
* all 2.5+.

Reported by @Onvember.

olegrok:
This patch removes io_to_fio hack in favour a new one. Previous
version used fio metatable and it was broken after
tarantool/tarantool@3d5b4da. This patch just set file descriptor
numer to stdout (2) or stderr (3) when it required. So it seems
enough to solve this problem.

An alternative way is to use `struct fio_handle` directly. But it
seems more danger way than proposed.

Closes #112

### DIFF
--- a/.github/workflows/push_rockspec.yaml
+++ b/.github/workflows/push_rockspec.yaml
@@ -1,0 +1,45 @@
+name: Push rockspec
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+
+env:
+  ROCK_NAME: "luatest"
+
+jobs:
+  push-scm-rockspec:
+    runs-on: [ ubuntu-latest ]
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - uses: actions/checkout@master
+
+      - uses: tarantool/rocks.tarantool.org/github-action@master
+        with:
+          auth: ${{ secrets.ROCKS_AUTH }}
+          files: ${{ env.ROCK_NAME }}-scm-1.rockspec
+
+  push-tagged-rockspec:
+    runs-on: [ ubuntu-latest ]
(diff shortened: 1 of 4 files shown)


rosik:
CI is still red, but with new error messages

olegrok:
@rosik 

This failures aren't related to my changes. I don't touch luatest code. I've just fixed a test.

Steap2448 (approved):

PR #135, opened 2021-06-02
Run tests in individual sandbox

### LINKED ISSUE #122: Sandbox tests
We lack the way to run tests in their personal sandboxes. Nowadays the flow sequence is approximately the following:

```lua
require('test1')
reqture('test2')

test1.before_all()
test1.test()
test1.after_all()

test2.before_all()
test2.test()
test2.after_all()
```

If the first test modifies global environment (applies mocks) it may affect subsequent tests in a rather puzzled way. See also https://github.com/tarantool/cartridge/pull/1285

Steap2448:
What has been done? Why? What problem is being solved?

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #122

### DIFF
--- a/luatest/runner.lua
+++ b/luatest/runner.lua
@@ -75,10 +75,11 @@ Options:
   -v, --verbose:          Increase verbosity
   -q, --quiet:            Set verbosity to minimum
   -c                      Disable capture
+  --sandbox               Sandbox group of tests
   -b                      Print full backtrace (don't remove luatest frames)
   -e, --error:            Stop on first error
   -f, --failure:          Stop on first failure or error
-  --shuffle VALUE:        Set execution order:
+  -s, --shuffle VALUE:        Set execution order:
                             - group[:seed] - shuffle tests within group
                             - all[:seed] - shuffle all tests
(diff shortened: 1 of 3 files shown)


Totktonada:
If you plan to going toward the 'clean tarantool state from inside tarantool' direction, I suggest to pay attention to our attempts to solve problems of the same kind in test-run:

* [pretest_clean.lua](https://github.com/tarantool/test-run/blob/b15160f430a51584c5ef943470a7d23e10354412/pretest_clean.lua) script.
* [Issues](https://github.com/tarantool/test-run/issues?q=is%3Aissue+is%3Aopen+pretest_clean) around the script.

However I'll warn: this is almost infinite road and the implementation always will be flawed.

The reliable way is to actually run a separate tarantool instance. This, OTOH, leads to its own subtle differences (another working directory at least).

----

I would think about usablity. Should the option be a user choise (command line option), a project wide configuration value (stored within the repository), a file/group/test scope configuration value? I don't know an answer, just highlighted that we should pay attention to this.

PR #136, opened 2021-06-04
Fix handling flaky tests

void234:
"luatest -r 2" etc would terminate with
"attempt to concatenate field 'test_iteration' (a nil value)"
if some test would fail not on the first iteration.

Broken in 940b4c1
(Tue Dec 3 11:32:59 2019 +0300).

Reproducer (launch with "-r 2"):
```
local t = require('luatest')

g_flaky = t.group('flaky')

local counter = 1
g_flaky.test_flaky = function()
    t.fail_if(counter > 1, 'Boo!')
    counter = counter + 1
end
```

I didn't forget about

- [x] Tests
- [ ] Changelog
- [ ] Documentation

### DIFF
--- a/luatest/runner.lua
+++ b/luatest/runner.lua
@@ -404,7 +404,7 @@ end
 function Runner.mt:invoke_test_function(test, iteration)
     local err = self:protected_call(test.group, test.method, test.name)
     if iteration > 1 and err.status ~= 'success' then
-        err.message = tostring(err.message) .. '\nIteration ' .. self.test_iteration
+        err.message = tostring(err.message) .. '\nIteration ' .. iteration
     end
     self:update_status(test, err)
 end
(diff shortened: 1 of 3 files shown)


rosik (commented):
Hi! Thanks for your patch. The fix is clear, but could you add a test for it?

void234:
> Hi! Thanks for your patch. The fix is clear, but could you add a test for it?

Done (commit message / PR description). Should I put this test somewhere else for automatic testing?

Steap2448:
> > Hi! Thanks for your patch. The fix is clear, but could you add a test for it?
> 
> Done (commit message / PR description). Should I put this test somewhere else for automatic testing?

Greetings! Thank you for your patch once again. I have added your test, think it will do fine.

Steap2448 (approved):

rosik (approved):

PR #138, opened 2021-06-15
Add line with tap version to the tap output

VitaliyaIoffe:
Tap protocol is used to the first line as a line with the version.


Steap2448 (approved):
Thanks for the patch!

PR #140, opened 2021-06-30
server: add restart function

VitaliyaIoffe:
Add restart function for server

Also, add test for check restart was correct.

### DIFF
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -127,6 +127,17 @@ function Server:start()
     log.debug('Started server PID: ' .. self.process.pid)
 end
 
+--- Restart server process.
+function Server:restart(args)
+    if not self.process then
+        log.warn("Process isn't running")
+    end
+    self:stop()
+    self.args = args or {}
+    self:start()
+    log.debug('Restarted server PID: ' .. self.process.pid)
+end
+
 --- Stop server process.
 function Server:stop()
     if self.net_box then
--- a/test/server_test.lua
+++ b/test/server_test.lua
@@ -51,6 +51,37 @@ g.test_start_stop = function()
     t.assert_equals(s.args, orig_args)
 end
 
+g.test_restart = function()
+--     Start server. Restart with same args. Check it was started. Check args
+    local workdir = fio.pathjoin(datadir, 'restart')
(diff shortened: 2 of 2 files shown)


luatest/server.lua
```diff
@@ -127,6 +127,19 @@ function Server:start()
     log.debug('Started server PID: ' .. self.process.pid)
 end
 
+--- Restart server process.
+function Server:restart(args)
+    if self.process then
+        self:stop()
+    else
+         log.warn('Try to restart, but server is not started')
+    end
```
Steap2448, line 136:
```suggestion
    if not self.process then
        log.warn('Process isn't running')
    end
    
    self:stop()
```

test/server_test.lua
```diff
@@ -51,6 +51,37 @@ g.test_start_stop = function()
     t.assert_equals(s.args, orig_args)
 end
 
+g.test_restart = function()
+--     Start server. Restart with same args. Check it was started. Check args
```
Steap2448, line 55: Let's split this comment into several little ones and place them near corresponding sections.

test/server_test.lua
```diff
@@ -51,6 +51,37 @@ g.test_start_stop = function()
...
+    end)
+    t.assert_equals(s.args, orig_args)
+
+    local new_args = {'test', 'args'}
+    s:restart(new_args)
+    pid = s.process.pid
+    t.helpers.retrying({timeout = 0.5}, function()
+        t.assert(Process.is_pid_alive(pid))
+    end)
+    s:stop()
```
Steap2448, line 78: Why are you stopping the server?

Steap2448 (changes requested):
Thanks for your patch! There are some comments but they are mostly about style.

test/server_test.lua
VitaliyaIoffe: Otherwise, it wasn't be stopped after the suite.
But I realized that checking args is much better after the restart.

VitaliyaIoffe: Otherwise, it wasn't be stopped after the suite.
But I realized that checking args is much better after the restart.

Steap2448 (approved):

PR #144, opened 2021-07-14
server: add convenient eval

VitaliyaIoffe:
Syntactic sugar for using net_box eval

### DIFF
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -213,6 +213,11 @@ function Server:eval(...)
     return self.net_box:eval(...)
 end
 
+function Server:call(...)
+    assert(self.net_box, 'net_box is not connected')
+    return self.net_box:call(...)
+end
+
 function Server:coverage(action)
     self:eval('require("luatest.coverage_utils").' .. action .. '()')
 end
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -208,8 +208,13 @@ function Server:http_request(method, path, options)
     return response
 end
 
+function Server:eval(...)
+    assert(self.net_box, 'net_box is not connected')
+    return self.net_box:eval(...)
+end
+
 function Server:coverage(action)
-    self.net_box:eval('require("luatest.coverage_utils").' .. action .. '()')
+    self:eval('require("luatest.coverage_utils").' .. action .. '()')
 end
 
 return Server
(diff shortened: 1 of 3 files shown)


luatest/server.lua
```diff
@@ -208,6 +208,10 @@ function Server:http_request(method, path, options)
     return response
 end
 
+function Server:eval(command)
```
Steap2448, line 211: Arguments and options should be passed as well. See doc [here](https://www.tarantool.io/en/doc/latest/reference/reference_lua/net_box/#lua-function.conn.eval).

luatest/server.lua
```diff
@@ -208,6 +208,10 @@ function Server:http_request(method, path, options)
...
+    return self.net_box:eval(command)
```
Steap2448, line 212:
```suggestion
function Server:eval(command)
    assert(self.net_box, 'net_box is not connected')
    return self.net_box:eval(command)
```

luatest/server.lua
```diff
@@ -208,6 +208,10 @@ function Server:http_request(method, path, options)
...
+end
+
```
Steap2448, line 214: Consider adding a corresponding function for `net_box:call`.

Steap2448 (changes requested):
Thanks for your patch! Consider the following comments.

luatest/server.lua
rosik:
```suggestion
function Server:eval(...)
    assert(self.net_box, 'net_box is not connected')
    return self.net_box:eval(...)

**Comment by rosik** [comment]
I like the idea. Please, make use of the new function:

```console
$ grep -Rn 'net_box:eval'
README.rst:280:    server.net_box:eval('return do_something(...)', {arg1, arg2})
luatest/server.lua:212:    self.net_box:eval('require("luatest.coverage_utils").' .. action .. '()')
test/server_test.lua:111:    t.assert_equals(server.net_box:eval('return os.getenv("custom_env")'), 'test_value')
```

luatest/server.lua
VitaliyaIoffe: Thank you! Fixed

Steap2448 (approved):
Nice one!

rosik (approved):

PR #164, opened 2021-09-07
Repeat _each / _test hooks

### LINKED ISSUE #161: Option repeat don't care about setup/teardown
I want to execute testcase for measuring performance for a number of times using option `-r N` and aggregate results from a test log using option `-c`. However luatest doesn't make a cleanup before next iteration and test faled. I propose to invoke setup (`before_each`) and teardown (`after_each`) when `-r` option is used.

https://github.com/tarantool/luatest/blob/master/luatest/runner.lua#L393-L402

Steap2448:
Repeat `_each` and `_test` hooks when `-r` option is specified.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] ~~Documentation~~

Close #161

### DIFF
--- a/test/hooks_test.lua
+++ b/test/hooks_test.lua
@@ -288,7 +288,7 @@ g.test_before_and_after_error = function()
         t2.after_test('test', function() error('custom-error') end)
         t2.after_test('test', function() table.insert(hooks, 'after_test3') end)
         t2.test = function() table.insert(hooks, 'test') end
-    end)
+    end, {'--repeat', '2'})
 
     t.assert_equals(result, 1)
     t.assert_equals(hooks, {'before_test1', 'after_test1'})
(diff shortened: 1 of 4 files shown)


test/hooks_test.lua
```diff
@@ -311,3 +311,25 @@ g.test_each_error = function()
...
+        t2.test_b = function() table.insert(hooks, 'test_b') end
+    end, {'--repeat', '2'})
+
+    t.assert_equals(result, 0)
+    t.assert_equals(hooks, {
+        'before_each', 'before_test', 'test_a', 'after_each',
+        'before_each', 'before_test', 'test_a', 'after_each',
+        'before_each', 'test_b', 'after_test', 'after_each',
+        'before_each', 'test_b', 'after_test', 'after_each',
+    })
```
rosik, line 334: Would you like to extend it with before_all and after_all?

Also, are there tests that cover:
- an error in `before_each` in conjunction with `--repeat` option
- a case when n-th test fails

Steap2448: > Would you like to extend it with before_all and after_all?

I don't think that repeating `_all` hooks is a good idea. They are supposed to be before and after **all** tests (evidently), but with repeat option it won't be this way.

rosik: I didn't mean calling `before_all` twice but extending the test with it. Never mind, you can merge it as is.

rosik (approved):

test/hooks_test.lua
Steap2448: My bad. Added it to the test

PR #170, opened 2021-09-22
Add XFail support

### LINKED ISSUE #158: Add XFail support
Sometimes we have a testcase blocked by bug. To keep CI green we need to temporarily disable it with, for example, `skip_if()` function. However to avoid forever disabled testcases one should periodically check bug and remove `skip_if` when bug has been resolved as fixed. These can be avoided with xfail status. Behaviour of XFail similar to skip, but testcase with xfail should always run. If testcase has failed then report XFail status with a reason, if testcase has passed then report FAIL status and remind to remove xfail in a testcase code.

Steap2448:
Test can be marked as xfail. It will be treated as failed when succeeded.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add `xfail` option
+
 ## 0.5.5
 
 - Repeat `_each` and `_test` hooks when `--repeat` is specified.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Add `xfail` option
+- Add `xfail` test marks support.
 
 ## 0.5.5
 
--- a/README.rst
+++ b/README.rst
@@ -237,6 +237,23 @@ List of luatest functions
 | ``group (name)``                                                   | Create group of tests.                        |
 +--------------------------------------------------------------------+-----------------------------------------------+
 
+.. _xfail
+
+---------------------------------
+XFail
+---------------------------------
+
+Test can be marked as xfail. It will be treated as failed when succeeded.
+
+.. code-block:: Lua
+
+    local g = t.group()
+
+    g.xfail['test_fail_1'] = true
+    g.xfail['test_fail_2'] = 'Must fail no matter what'
+    g.test_fail_1 = function() t.assert_equals(1, 2) end
+    g.test_fail_2 = function() t.assert_equals(3, 4) end
+
 .. _capturing-output:
 
 ---------------------------------
--- a/README.rst
+++ b/README.rst
@@ -219,6 +219,10 @@ List of luatest functions
 | ``fail_if (condition, message)``                                   | Stops a test due to a failure if condition    |
 |                                                                    | is met.                                       |
 +--------------------------------------------------------------------+-----------------------------------------------+
+| ``xfail (message)``                                                | Mark test as xfail.                           |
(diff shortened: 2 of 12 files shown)


ligurio:
AFAIU it is not possible to specify a reason of XFail and show that reason near the test status in output. Am I right?

Steap2448:
> AFAIU it is not possible to specify a reason of XFail and show that reason near the test status in output. Am I right?

It is possible now.

Steap2448:
> AFAIU it is not possible to specify a reason of XFail and show that reason near the test status in output. Am I right?

It is possible now.

luatest/assertions.lua
```diff
@@ -15,6 +15,16 @@ local prettystr_pairs = pp.tostring_pair
...
+local mt = {}
+setmetatable(M, mt)
+
+mt.xfail = false
+function mt.is_xfail()
+    local xfail = mt.xfail
+    mt.xfail = false
+    return xfail
+end
+
```
rosik, line 27: I don't like the idea of using metatables as private storage. Look, there's `M.private` definition two lines below. Let's put the `is_xfail` function there?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add `xfail` option
```
rosik, line 5:
```suggestion
- Add `xfail` test marks support.
```

README.rst
```diff
@@ -219,6 +219,10 @@ List of luatest functions
 | ``fail_if (condition, message)``                                   | Stops a test due to a failure if condition    |
 |                                                                    | is met.                                       |
 +--------------------------------------------------------------------+-----------------------------------------------+
+| ``xfail (message)``                                                | Mark test as xfail.                           |
++--------------------------------------------------------------------+-----------------------------------------------+
+| ``xfail_if (condition, message)``                                  | Mark test as xfail if condition is met.       |
++--------------------------------------------------------------------+-----------------------------------------------+
```
rosik, line 225:
```suggestion
| ``xfail (message)``                                                | Mark test as :ref:`XFail`.                    |
+--------------------------------------------------------------------+-----------------------------------------------+
| ``xfail_if (condition, message)``                                  | Mark test as :ref:`XFail` if condition is met.|
+--------------------------------------------------------------------+-----------------------------------------------+
```

README.rst
```diff
@@ -237,6 +241,24 @@ List of luatest functions
...
+
+.. code-block:: Lua
+
+    local g = t.group()
+
+    g.test_fail = function()
+        t.xfail('Must fail no matter what')
+        t.assert_equals(3, 4)
+    end
+
```
rosik, line 261:
```suggestion
.. _XFail

---------------------------------
XFail
---------------------------------

The ``xfail`` mark makes test results to be interpreted vice versa: it's
threated as passed when an assertion fails, and it fails if no errors are
raised. It allows one to mark a test as temporarily broken due to a bug in some
other component which can't be fixed immediately. It's also a good practice to
keep xfail tests in sync with an issue tracker.

.. code-block:: Lua

    local g = t.group()

    g.test_fail = function()
        t.xfail('Must fail no matter what')
        t.assert_equals(3, 4)
    end

XFail only applies to the errors raised by the luatest assertions. Regular Lua
errors still cause the test failure.

```

rosik (approved):

PR #207, opened 2021-12-14
Fix confusing err message from `assert_not_equals`

### LINKED ISSUE #206: assert_not_equals prints confusing message
```lua
t.assert_not_equals(a, b)
```
Assert message: "Received unexpected value: 0"
It would be nice to see the values of `a` and `b`.

ylobankov:
The `assert_not_equals` function printed the following error message
when the values were equal:

    Received unexpected value: <value>

This message sounds quite confusing because in this case the question
is raised: 'what is the expected value?'. But the expected value is any
value different from `<value>`. So the error message was changed to

    Actual and expected values are equal: <value>

which is more comprehensible.

Closes #206

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-- Fix invalid arguments logging in some of assertions.
+- Fix invalid arguments logging in some assertions.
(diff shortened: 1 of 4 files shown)


yngvar-antonsson (approved):

Totktonada (approved):
LGTM.

PR #209, opened 2021-12-16
Check hooks are defined only for existing tests

### LINKED ISSUE #205: Check if before_test/after_test doesn't belong to a test
Now if you create a test hook, that doesn't belong to any test (e.g. if somebody makes a typo in test_name), you won't know that.
I propose to warn if there are hooks without existing tests. 

```lua
g.test_name_without_typo = function() end
g.before_test('test_name_with_typo', function() end)
```

ylobankov:
When a named test hook, that doesn't belong to any test (for example, if
someone makes a typo in test name), was being defined, it was impossible
to know about that. This might lead to unnecessary debugging actions
from the developer's side to understand what's going on and why the hook
is not running.

Now to catch such cases luatest checks that hooks are defined only for
existing tests and raises an error with a message like that:

    There is no test with name '<test name>' but hook 'before_test' is 
    defined for it

which should give glue to developers what is wrong.

Closes #205

### DIFF
--- a/luatest/hooks.lua
+++ b/luatest/hooks.lua
@@ -183,6 +183,17 @@ function export.patch_runner(Runner)
     -- Run group hook and save possible error to the group object.
     utils.patch(Runner.mt, 'start_group', function(super) return function(self, group)
         super(self, group)
+        -- Check while starting group that 'before_test' and 'after_test' hooks are defined only for existing tests.
+        for _, hooks_type in ipairs({'before_test', 'after_test'}) do
+            for full_test_name in pairs(group[hooks_type .. '_hooks']) do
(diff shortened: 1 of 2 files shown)


yngvar-antonsson:
Why don't you check an error "There is no test with name '%s' but hook '%s' is defined for it" in tests?

yngvar-antonsson (approved):

ylobankov:
> Why don't you check an error "There is no test with name '%s' but hook '%s' is defined for it" in tests?

Now checking.

Totktonada (approved):
I like the idea. I don't know this code well, but the logic seems correct. No objections from me.

PR #218, opened 2022-01-24
Print message instead of `nil` value

### LINKED ISSUE #171: luatest prints nil when message in assertion is empty
When testcase contains a `skip_if` without an optional message luatest prints `nil` after test status:
```
sergeyb@pony:~/sources/MRG/expirationd$ ./.rocks/bin/luatest -v test/unit/start_key_test.lua 
Started on Wed Sep 22 17:51:26 2021
    start_key.engine:"memtx".index_type:"BITSET".test_passing ... (0.010s) skip
nil
    start_key.engine:"memtx".index_type:"BITSET".test_tree_index ... (0.009s) skip
...eyb/sources/MRG/expirationd/test/unit/start_key_test.lua:73: Unsupported index type
    start_key.engine:"memtx".index_type:"BITSET".test_hash_index ... (0.010s) skip
...eyb/sources/MRG/expirationd/test/unit/start_key_test.lua:130: Unsupported index type
    start_key.engine:"memtx".index_type:"HASH".test_passing ... (0.011s) skip
nil
    start_key.engine:"memtx".index_type:"HASH".test_tree_index ... (0.012s) skip
```

luatest 0.5.5

ylobankov:
When a test case contained the skip/skip_if/run_only_if function without
an optional message, luatest printed `nil` after the test status.

Now luatest prints the `(no reason specified)` message.

Closes #171

### DIFF
--- a/luatest/assertions.lua
+++ b/luatest/assertions.lua
@@ -118,7 +118,7 @@ end
 --
 -- @string message
 function M.skip(message)
-    utils.luatest_error('skip', message, 2)
+    utils.luatest_error('skip', message or '(no reason specified)', 2)
 end
 
 --- Skip a running test if condition is met.
@@ -127,14 +127,14 @@ end
 -- @string message
 function M.skip_if(condition, message)
     if condition and condition ~= nil then
-        utils.luatest_error('skip', message, 2)
+        utils.luatest_error('skip', message or '(no reason specified)', 2)
     end
 end
 
 function M.run_only_if(condition, message)
     -- continue a running test if condition is met, else skip it
     if not (condition and condition ~= nil) then
-        utils.luatest_error('skip', prettystr(message), 2)
+        utils.luatest_error('skip', message or '(no reason specified)', 2)
     end
 end
 
--- a/test/luatest_test.lua
+++ b/test/luatest_test.lua
@@ -79,6 +79,24 @@ g.test_skip_if_tnt_specific = function()
     t.assert_equals(helper.assert_failure(t.skip_if, {}, 'expected').status, 'skip')
 end
 
+g.test_skip_no_reason_specified = function()
+    local result = helper.assert_failure(t.skip)
+    t.assert_equals(result.status, 'skip')
+    t.assert_equals(result.message, '(no reason specified)')
+end
+
+g.test_skip_if_no_reason_specified = function()
+    local result = helper.assert_failure(t.skip_if, true)
+    t.assert_equals(result.status, 'skip')
+    t.assert_equals(result.message, '(no reason specified)')
+end
+
+g.test_run_only_if_no_reason_specified = function()
+    local result = helper.assert_failure(t.run_only_if, false)
+    t.assert_equals(result.status, 'skip')
+    t.assert_equals(result.message, '(no reason specified)')
+end
+
 g.test_success_if_tnt_specific = function()
     assert_any_error(t.success_if, box.NULL)
     t.assert_equals(helper.assert_failure(t.success_if, true).status, 'success')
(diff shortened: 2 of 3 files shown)


test/luatest_test.lua
```diff
@@ -79,13 +79,31 @@ g.test_skip_if_tnt_specific = function()
...
+end
+
 g.test_success_if_tnt_specific = function()
     assert_any_error(t.success_if, box.NULL)
     t.assert_equals(helper.assert_failure(t.success_if, true).status, 'success')
     t.assert_equals(helper.assert_failure(t.success_if, {}).status, 'success')
 end
 
-g.test_assert_aliases = function ()
+g.test_assert_aliases = function()
```
artembo, line 106: It seems the formatting fix is not related to the task, I would suggest discarding this change to make the change history clean.

luatest/assertions.lua
```diff
@@ -127,14 +127,14 @@ end
...
-        utils.luatest_error('skip', message, 2)
+        utils.luatest_error('skip', message or 'No reason', 2)
     end
 end
 
 function M.run_only_if(condition, message)
     -- continue a running test if condition is met, else skip it
     if not (condition and condition ~= nil) then
-        utils.luatest_error('skip', prettystr(message), 2)
+        utils.luatest_error('skip', message or 'No reason', 2)
```
artembo, line 137: Was: `prettystr(message)`
Become: `message or 'No reason'` — should it be `prettystr(message) or 'No reason'`?

artembo (changes requested):

luatest/assertions.lua
ylobankov: IMHO, no `prettystr` should be here. If you look into the assertions.lua file, you will see that `prettystr` is used in asserting functions to make pretty error messages. The `skip` and `skip_if` functions don't use it, so the `run_only_if` function shouldn't as well.

artembo: or like `prettystr(message or 'No reason')`

luatest/assertions.lua
```diff
@@ -118,7 +118,7 @@ end
 --
 -- @string message
 function M.skip(message)
-    utils.luatest_error('skip', message, 2)
+    utils.luatest_error('skip', message or 'No reason', 2)
```
NickVolynkin, line 121:
```suggestion
    utils.luatest_error('skip', message or '(no reason specified)', 2)
```

There's a slight difference:

* This test was skipped for no reason
* This test was skipped, but no reason was specified

luatest/assertions.lua
```diff
@@ -127,14 +127,14 @@ end
 -- @string message
 function M.skip_if(condition, message)
     if condition and condition ~= nil then
-        utils.luatest_error('skip', message, 2)
+        utils.luatest_error('skip', message or 'No reason', 2)
```
NickVolynkin, line 130: Hey, but we know the reason here. It's because `condition` was true. 

```suggestion
        utils.luatest_error('skip', message or '(skip condition was true, no message specified)', 2)
```

luatest/assertions.lua
ylobankov: If we use the variant with `prettystr`, we will get message wrapped by `"` which looks ugly.

```
$ ./.rocks/bin/luatest -v test/unit/start_key_test.lua
Started on Tue Jan 25 11:50:14 2022
    start_key.engine:"memtx".index_type:"HASH".test_passing ... (0.002s) skip
...ankov/Workspace/expirationd/test/unit/start_key_test.lua:27: "(no reason specified)"
```

IMHO, we don't need `prettystr` at all.

test/luatest_test.lua
ylobankov: Ok, fixed.

luatest/assertions.lua
ylobankov: Fixed.

luatest/assertions.lua
ylobankov: IMHO, it's redundant. Let's have just `(no reason specified)` which is more than enough.

NickVolynkin (approved):

PR #219, opened 2022-01-25
Raise error if net_box_uri > max unix socket path

### LINKED ISSUE #152: Raise error if length of self.net_box_uri > UNIX_PATH_MAX
Cause now luatest supports sockets, we need to check that socket's path length is less than `UNIX_PATH_MAX` in `server.lua`.

ylobankov:
Since luatest supports in luatest/server.lua the socket connection to
tarantool instances via the net_box_uri param, we need to check that
the socket's path length is less than the max length of the Unix domain
socket path allowed by a system. 
Otherwise, we will get the `No buffer space available` error which is not 
clear for users and may take some time to understand what's wrong.

Closes #152

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix confusing error message from `assert_items_include` function.
 - Print `(no reason specified)` message instead of `nil` value when the test is 
   skipped and no reason is specified.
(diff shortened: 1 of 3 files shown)


artembo (approved):
Looks good to me.

NickVolynkin (approved):

Issue #238, opened 2022-07-26
Move luatest helpers from misc.lua to luatest repository


NickVolynkin:
In [tarantool](https://github.com/tarantool/tarantool) repo we have some luatest helpers and it would be nice to move them to this repo to share helpers/utils around the projects using luatest. 

This issue is about moving helpers from misc.lua.

---

Related pull requests:

- https://github.com/tarantool/luatest/pull/247
- https://github.com/tarantool/test-run/pull/354
- https://github.com/tarantool/tarantool/pull/7857

PR #247, opened 2022-10-07
Add new module tarantool.lua

ylobankov:
This patch adds a new module tarantool.lua to the repo that is a copy of
the test/luatest_helpers/misc.lua module [1] in the tarantool/tarantool
repo. The later should be deleted in the nearest future.

[1] https://github.com/tarantool/tarantool/blob/5c62f01bb6d7a88855108925b8ed0ba3ca344d2b/test/luatest_helpers/misc.lua

Part of #238

### DIFF
--- a/luatest/init.lua
+++ b/luatest/init.lua
@@ -5,6 +5,11 @@ local luatest = setmetatable({}, {__index = require('luatest.assertions')})
 luatest.Process = require('luatest.process')
 luatest.VERSION = require('luatest.VERSION')
 
+--- Tarantool related functions.
+--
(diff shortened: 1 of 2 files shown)


locker (approved):
*Technically* this patch looks correct to me so ack **however** I don't understand why you need to move the helpers to the luatest repository. Since these helpers are very specific to the Tarantool project, I don't think they can be reused in anywhere else. At the same time, it'll surely complicate adding new tarantool-specific helpers or patching the old ones.  I just hope that you know what you're doing...

PR #255, opened 2022-10-25
Add new module replica_proxy.lua

ylobankov:
This patch adds a new module replica_proxy.lua to the repo that is
a copy of the test/luatest_helpers/proxy/proxy.lua module [1] in the
tarantool/tarantool repo. The later should be deleted in the nearest
future.

The new module is intended to manage connection to Tarantool replication
instances via proxy.

Also, the replica_conn.lua module is added that is considered as an
internal one and not exposed to public API.

[1] https://github.com/tarantool/tarantool/blob/ad420846861a117b2b7e82b434acd1372c6abe08/test/luatest_helpers/proxy/proxy.lua

Part of https://github.com/tarantool/luatest/issues/251

### DIFF
--- a/luatest/replica_proxy.lua
+++ b/luatest/replica_proxy.lua
@@ -1,0 +1,139 @@
+--- Manage connection to Tarantool replication instances via proxy.
+--
+-- @module luatest.replica_proxy
+
+local checks = require('checks')
+local fiber = require('fiber')
+local log = require('log')
+local socket = require('socket')
+
+local Connection = require('luatest.replica_conn')
+
+local TIMEOUT = 0.001
+local BACKLOG = 512
+
+local Proxy = {
+    constructor_checks = {
+        client_socket_path = 'string',
+        server_socket_path = 'string',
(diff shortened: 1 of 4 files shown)


grafin (approved):
Thanks for the patch! Just one minor comment from me, otherwise LGTM.

luatest/replica_proxy.lua
```diff
@@ -0,0 +1,137 @@
...
+    self.__index = self
+    return object
+end
+
+--- Build proxy object.
+--
+-- @param object
+-- @string object.client_socket_path Path to client Unix socket.
+-- @string object.server_socket_path Path to server Unix socket.
+-- @tab[opt] object.process_client TBD.
```
grafin, line 35: process_client and process_server description can be found in https://github.com/tarantool/tarantool/commit/b907e7260920b05886630d1a4e057d865367b403 commit message.

ylobankov: Thanks for pointing it out.

PR #257, opened 2022-11-01
Print Tarantool version used by luatest 

### LINKED ISSUE #240: Print Tarantool version used by luatest
It would be useful to print a version of tarantool used for running tests.

Dirty patch (probably patch is not complete, because luatest can use two tarantool binaries: for running luatest itself and for running tests):
```diff
diff --git a/bin/luatest b/bin/luatest
index ada847d..3479aed 100755
--- a/bin/luatest
+++ b/bin/luatest
@@ -1,3 +1,4 @@
 #!/usr/bin/env tarantool
 
+print(require("tarantool").version)
 require('luatest.cli_entrypoint')()
```

```sh
sergeyb@pony:~/sources/MRG/tarantool$ luatest --verbose test/app-luatest/http_client_test.lua                                               
2.11.0-entrypoint-355-g43d267b97                                                                                                            
Running with --shuffle group:1067                                                                                                           
Started on Mon Aug  8 14:39:21 2022 
...
```

slavakirichenko:
Printing out tarantool version at bin/luatest.

Added method to printing out used tarantool version.

Resolves #240

### DIFF
--- a/bin/luatest
+++ b/bin/luatest
@@ -1,3 +1,5 @@
 #!/usr/bin/env tarantool
 
+print(('Tarantool version is %s'):format(require('tarantool').version))
+
 require('luatest.cli_entrypoint')()


bin/luatest
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env tarantool
+print(require("tarantool").version)
```
ylobankov, line 2: 1. Could you add a blank line between the shebang and print?
2. Could you please use `'` instead of `"`?
3. I think we need to highlight this somehow. For example, 

```lua
print(('Tarntool version is %s'):format(require('tarantool').version))
```

ylobankov (commented):
You need to swap the commit subject and description.

```
Printing out tarantool version at bin/luatest

Resolves #240
```

bin/luatest
slavakirichenko: All is done.

PR #261, opened 2022-11-16
Forbid negative values for -r,--repeat option

### LINKED ISSUE #250: Forbid negative options for repeat
```bash
luatest test/fixtures/flaky_group.lua -r -10
Running with --shuffle group:1463
..
Ran 2 tests in 0.000 seconds, 2 succeeded, 0 failed
```

ylobankov:
Also, forbid zero as it doesn't make sense for this option.

Fixes #250


yngvar-antonsson (approved):

test/luaunit/utility_test.lua
```diff
@@ -580,7 +580,10 @@ function g.test_parse_cmd_line()
     -- repeat
     assert_subject({'--repeat', '123'}, {exe_repeat=123})
     assert_subject({'-r', '123'}, {exe_repeat=123})
-    t.assert_error_msg_contains('Invalid value for -r option. Integer required', subject, {'-r', 'bad'})
+    t.assert_error_msg_contains('Invalid value for --repeat option. Positive integer required',
```
NickVolynkin, line 583: tests don't check negative values :)

ylobankov: Actually, it is not necessary because zero already covers this case.

slavakirichenko (approved):

Issue #239, opened 2022-07-26
Move luatest helpers from server.lua to luatest repository


NickVolynkin:
In [tarantool](https://github.com/tarantool/tarantool) repo we have some luatest helpers and it would be nice to move them to this repo to share helpers/utils around the projects using luatest. 

This issue is about moving helpers from server.lua.

PR #266, opened 2022-11-28
Merge server_v2.lua into server.lua

ylobankov:
It was decided to merge module server_v2.lua into server.lua to have one
common interface for managing Tarantool instances. The new server.lua is
fully compatible with the old version, which means that nothing changed
for the projects using luatest previously.

In the same time, the new server.lua has one significant feature that
differs it from the old version: by default, it waits until a server is
ready if no custom executable path to start the server is specified.
Otherwise, we fall back to the old behavior.

Part of #239

---
Documentation preview: https://docs.d.tarantool.io/en/doc/luatest/reference/reference_rock/luatest/classes/luatest.server/

(deployed by https://github.com/tarantool/doc/actions/runs/3592100174)


luatest/server.lua
```diff
@@ -36,47 +49,109 @@ local Server = {
...
-        end
+
+    if self.net_box_uri == nil and self.net_box_port == nil then
+        self.net_box_uri = self.build_listen_uri(self.alias)
+        fio.mktree(self.vardir)
     end
     if self.net_box_uri == nil and self.net_box_port then
         self.net_box_uri = 'localhost:' .. self.net_box_port
     end
+    if uri.parse(self.net_box_uri).host == 'unix/' then
```
NickVolynkin, line 141: when is it not unix?
when is it unix, but not linux? with macos?

luatest/server.lua
```diff
@@ -1,29 +1,42 @@
...
+    fio.dirname(fio.abspath(package.search('luatest.server'))), 'server_instance.lua'
+)
+local WAIT_TIMEOUT = 60
+local WAIT_DELAY = 0.1
 
 local Server = {
     constructor_checks = {
-        command = 'string',
-        workdir = 'string',
+        command = '?string',
```
NickVolynkin, line 33: Order of variables here does not match the order in the constructor docstrings, lines 63-94

luatest/server.lua
```diff
@@ -36,47 +49,109 @@ local Server = {
...
+-- @return table
 function Server:new(object)
     checks('table', self.constructor_checks)
     self:inherit(object)
     object:initialize()
     return object
 end
 
+-- Initialize the server object.
 function Server:initialize()
```
NickVolynkin, line 103: It's not a `local` function, but the docstring does not start with `---`.

luatest/server.lua
```diff
@@ -106,21 +181,53 @@ function Server:initialize()
...
--- @return map
+--- Create a table with env variables based on the constructor params.
+-- The result will be passed into the server process.
+-- Table consists of the following entries:
+--
+--   * TARANTOOL_WORKDIR
+--   * TARANTOOL_LISTEN
+--   * TARANTOOL_ALIAS
+--   * TARANTOOL_HTTP_PORT
+--   * TARANTOOL_BOX_CFG
```
NickVolynkin, line 192:
```suggestion
--   * `TARANTOOL_WORKDIR`
--   * `TARANTOOL_LISTEN`
--   * `TARANTOOL_ALIAS`
--   * `TARANTOOL_HTTP_PORT`
--   * `TARANTOOL_BOX_CFG`
```

luatest/server.lua
```diff
@@ -106,21 +181,53 @@ function Server:initialize()
...
 function Server:build_env()
-    return {
+    local res = {
         TARANTOOL_WORKDIR = self.workdir,
-        TARANTOOL_HTTP_PORT = self.http_port,
         TARANTOOL_LISTEN = self.net_box_port or self.net_box_uri,
         TARANTOOL_ALIAS = self.alias,
     }
+    if self.http_port ~= nil then
+        res.TARANTOOL_HTTP_PORT = self.http_port
```
NickVolynkin, line 202: And if `self.http_port` is `nil`, then `res.TARANTOOL_HTTP_PORT` is also `nil`, right? So anyway it equals `self.http_port`.

luatest/server.lua
```diff
@@ -106,21 +181,53 @@ function Server:initialize()
...
+-- For now, only UNIX sockets are supported.
+--
+-- @string server_alias Server alias.
+-- @return string
+function Server.build_listen_uri(server_alias)
+    return fio.pathjoin(Server.vardir, server_alias .. '.sock')
+end
+
+--- Start a server.
+-- Optionally waits until the server is ready.
```
NickVolynkin, line 220: If the default behavior is to wait, it is more correct to say that it optionally doesn't wait.

luatest/server.lua
ylobankov: It is about UNIX sockets, not operating system name.

luatest/server.lua
```diff
@@ -158,7 +284,8 @@ function Server:restart(params)
         alias = '?string',
 
         coverage_report = '?boolean',
-    })
+    }, {wait_until_ready = '?boolean'})
```
NickVolynkin, line 287: Why isn't this `wait_until_ready` a part of `params`? Also, the trailing `)` should probably be on a separate line.

luatest/server.lua
ylobankov: It is kind of an internal function. It shouldn't be invoked as `server:initialize()`, so I didn't document it.

BTW, the same should be for `server:build_env()`. This function is supposed to be an internal one as well.

luatest/server.lua
```diff
@@ -168,11 +295,28 @@ function Server:restart(params)
...
+-- Wait until the given condition is `true` (anything except `false` and `nil`).
+-- Throws an error when timeout exceeds.
+local function wait_for_condition(cond_desc, server, func, ...)
+    local deadline = clock.time() + WAIT_TIMEOUT
+    while true do
+        if func(...) then
+            return
+        end
+        if clock.time() > deadline then
+            error(('Timed out to wait for "%s" condition for server (alias: %s, workdir: %s, pid: %d) within %ds')
```
NickVolynkin, line 311: Maybe a better order is this: alias, pid, workdir. The workdir can be a long string.

luatest/server.lua

config.ld
```diff
@@ -6,7 +6,6 @@ file = {
     'luatest/http_response.lua',
     'luatest/runner.lua',
     'luatest/server.lua',
-    'luatest/server_v2.lua',
```
NickVolynkin, line 9: Please update the changelog

luatest/server.lua

luatest/server.lua
ylobankov: The default behavior is to wait unless a custom executable was provided while building the server object.

luatest/server.lua
ylobankov: I decided not to do this because it can be considered as not relevant changes.

luatest/server.lua
ylobankov: Because these are different entities. 

`params` - New parameters to restart the server with.
`opts` - Options for the `start()` function that is inside the `restart()` function.

luatest/server.lua
ylobankov: Actually, it is not long because of `basename`.

luatest/server.lua
NickVolynkin: Then why aren't both functions `local`?

config.ld
ylobankov: Ok, will do.

luatest/server.lua
NickVolynkin: Oh, I get it.

luatest/server.lua
NickVolynkin: What if I explicitly pass `true` and a custom executable?

luatest/server.lua
ylobankov: Then 'inheritance' will not work correctly.

luatest/server.lua
ylobankov: > What if I explicitly pass true and a custom executable?

The waiting will be on.

> Will the fact that there's a custom executable override the true value and make it not wait?

Yes.

NickVolynkin (approved):

Issue #269, opened 2022-12-07
luatest installed from tarantool rocks fails to run tarantool tests


locker:
Tarantool version: 2.11.0-entrypoint-754-g2ed06651c317
Tarantool luatest version: v0.5.7

Tarantool luatest installed from rocks using

```
tarantoolctl rocks install luatest
```

fails to run Tarantool functional tests:

[run from the root Tarantool repository directory]

```
$ luatest -v test/box-luatest/box_read_ffi_disable_test.lua
Tarantool version is 2.11.0-entrypoint-755-g39b65fa2fc40
Running with --shuffle group:2541
Started on Wed Dec  7 13:17:53 2022
    box-luatest.box_read_ffi_disable.disable_ffi:false.test_select ... (0.000s) error
/home/vlad/.rocks/share/tarantool/luatest/server.lua:309: Timed out to wait for "server is ready" condition for server (alias: master, workdir: master-fi5m070hl4LF, pid: 2267803) within 60s
    box-luatest.box_read_ffi_disable.disable_ffi:false.test_get ... (0.000s) error
/home/vlad/.rocks/share/tarantool/luatest/server.lua:309: Timed out to wait for "server is ready" condition for server (alias: master, workdir: master-fi5m070hl4LF, pid: 2267803) within 60s
    box-luatest.box_read_ffi_disable.disable_ffi:false.test_max ... (0.000s) error
/home/vlad/.rocks/share/tarantool/luatest/server.lua:309: Timed out to wait for "server is ready" condition for server (alias: master, workdir: master-fi5m070hl4LF, pid: 2267803) within 60s
    box-luatest.box_read_ffi_disable.disable_ffi:false.test_min ... (0.000s) error
/home/vlad/.rocks/share/tarantool/luatest/server.lua:309: Timed out to wait for "server is ready" condition for server (alias: master, workdir: master-fi5m070hl4LF, pid: 2267803) within 60s
    box-luatest.box_read_ffi_disable.disable_ffi:false.test_pairs ... (0.000s) error
/home/vlad/.rocks/share/tarantool/luatest/server.lua:309: Timed out to wait for "server is ready" condition for server (alias: master, workdir: master-fi5m070hl4LF, pid: 2267803) within 60s
    box-luatest.box_read_ffi_disable.disable_ffi:false.test_random ... (0.000s) error
[...]

PR #270, opened 2022-12-08
Fix instance file permissions when installing luatest as rock

ylobankov:
When luatest is installed as a rock, the internal server_instance.lua
script won't have execution permissions even though it has them in the
source tree, and won't be able to be run while a server start.

Let's start a server process as

    /abs/path/to/tarantool /abs/path/to/luatest/server_instance.lua

instead of just

    /abs/path/to/luatest/server_instance.lua

to bypass the problem.

Also, the shebang string and x-bit are removed from server_instance.lua.

Part of #269

### DIFF
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -232,13 +232,21 @@ function Server:start(opts)
         table.insert(log_cmd, string.format('%s=%q', k, v))
         env[k] = v
     end
+    table.insert(log_cmd, arg[-1])
     table.insert(log_cmd, self.command)
     for _, v in ipairs(self.args) do
         table.insert(log_cmd, string.format('%q', v))
     end
     log.debug(table.concat(log_cmd, ' '))
 
-    self.process = Process:start(self.command, self.args, env, {
+    -- When luatest is installed as a rock, server_instance.lua script won't
+    -- have execution permissions even though it has them in the source tree,
+    -- and won't be able to be run while a server start.
+    -- To bypass this issue, we start a server process as `tarantool <script>`
+    -- instead of just `<script>`.
+    local args = table.copy(self.args)
+    table.insert(args, 1, self.command)
+    self.process = Process:start(arg[-1], args, env, {
         chdir = self.chdir,
         output_prefix = self.alias,
     })
--- a/luatest/server_instance.lua
+++ b/luatest/server_instance.lua
@@ -1,5 +1,3 @@
-#!/usr/bin/env tarantool
-
 local fio = require('fio')
 local fun = require('fun')
 local json = require('json')


Totktonada:
Who requires an instance file to be an executable? I don't think that it is correct.

CMakeLists.txt
```diff
@@ -97,5 +97,6 @@ install(
 
 install(
   PROGRAMS ${CMAKE_CURRENT_SOURCE_DIR}/bin/luatest
+           ${CMAKE_CURRENT_SOURCE_DIR}/bin/luatest-tarantool-instance
```
locker, line 100: I agree with @Totktonada that an instance file shouldn't be executable. Can you run it with tarantool instead of executing it?

ylobankov:
> Who requires an instance file to be an executable? I don't think that it is correct.

Looks like [Server:start()](https://github.com/tarantool/luatest/blob/cb28d9389249df29f4164540152185c927822607/luatest/server.lua#L241) needs it.

CMakeLists.txt
ylobankov: Yes, I can, but ...

Original luatest code in the `Server:start()` function:
```lua
    self.process = Process:start(self.command, self.args, env, {
        chdir = self.chdir,
        output_prefix = self.alias,
    })
```

I can change it to something like this:
```lua
    local args = table.copy(self.args)
    table.insert(args, 1, self.command)
    self.process = Process:start('tarantool', args, env, {
        chdir = self.chdir,
        output_prefix = self.alias,
    })
```

The problem is that I have to provide the full executable path to make [execve](https://github.com/tarantool/luatest/blob/cb28d9389249df29f4164540152185c927822607/luatest/process.lua#L69) not fail with `No such file or directory` error.

Totktonada: Something like this should do the trick:

```lua
    local args = table.copy(self.args)
    table.insert(args, 1, self.command)
    self.process = Process:start(arg[-1], self.args, env, { <...> })
```

locker: > The problem is that I have to provide the full executable path to make [execve](https://github.com/tarantool/luatest/blob/cb28d9389249df29f4164540152185c927822607/luatest/process.lua#L69) not fail with `No such file or directory` error.

You can execute `/usr/bin/env tarantool ${ARGS}`.

ylobankov: > Something like this should do the trick:
> 
> ```lua
>     local args = table.copy(self.args)
>     table.insert(args, 1, self.command)
>     self.process = Process:start(arg[-1], self.args, env, { <...> })
> ```

Excellent, thanks!

ylobankov: > > The problem is that I have to provide the full executable path to make [execve](https://github.com/tarantool/luatest/blob/cb28d9389249df29f4164540152185c927822607/luatest/process.lua#L69) not fail with `No such file or directory` error.
> 
> You can execute `/usr/bin/env tarantool ${ARGS}`.

Thanks, I think @Totktonada's solution is exactly what I need here. Let's focus on it.

ylobankov: Fixed.

Totktonada (approved):

luatest/server.lua
```diff
@@ -232,13 +232,20 @@ function Server:start(opts)
...
         env[k] = v
     end
+    table.insert(log_cmd, arg[-1])
     table.insert(log_cmd, self.command)
     for _, v in ipairs(self.args) do
         table.insert(log_cmd, string.format('%q', v))
     end
     log.debug(table.concat(log_cmd, ' '))
 
-    self.process = Process:start(self.command, self.args, env, {
```
locker, line 241: Let's also remove the X bit and the sha-bang (`#!/usr/bin/env tarantool`) from `luatest/server_instance.lua`?

ylobankov: Fixed.

locker (approved):

Issue #237, opened 2022-07-25
Move luatest helpers from cluster.lua to luatest/replica_set.lua


ylobankov:
In [tarantool](https://github.com/tarantool/tarantool) repo we have some luatest helpers and it would be nice to move them to this repo to share helpers/utils around the projects using luatest. 

This issue is about moving helpers from cluster.lua.

PR #271, opened 2022-12-14
Add new module replica_set.lua

ylobankov:
This patch adds a new module replica_set.lua to the repo that is an
almost exact copy of the test/luatest_helpers/cluster.lua module [1]
in the tarantool/tarantool repo. The later should be deleted in the
nearest future.

[1] https://github.com/tarantool/tarantool/blob/943f7d1/test/luatest_helpers/cluster.lua

Part of #237

### DIFF
--- a/config.ld
+++ b/config.ld
@@ -6,9 +6,10 @@ file = {
     'luatest/http_response.lua',
     'luatest/runner.lua',
     'luatest/server.lua',
+    'luatest/replica_set.lua',
 }
 topics = {
     'CHANGELOG.md',
 }
 format = 'markdown'
-sort = true
+sort = 'true'
--- a/luatest/replica_set.lua
+++ b/luatest/replica_set.lua
@@ -1,0 +1,230 @@
+--- Class to manage groups of Tarantool instances with the same data set.
+--
+-- @classmod luatest.replica_set
+
+local checks = require('checks')
+local log = require('log')
+
+local helpers = require('luatest.helpers')
+local Server = require('luatest.server')
+
+local ReplicaSet = {}
+
+function ReplicaSet:inherit(object)
+    setmetatable(object, self)
+    self.__index = self
+    return object
+end
+
+--- Build a replica set object.
+--
+-- @tab[opt] object Table with the entries listed below. (optional)
+-- @tab[opt] object.servers List of server configurations to build server
+--   objects from and add them to the new replica set. See an example below.
+-- @return table
+-- @see luatest.server:new
+-- @usage
+--   local ReplicaSet = require('luatest.replica_set')
+--   local Server = require('luatest.server')
+--   local box_cfg = {
+--       replication_timeout = 0.1,
+--       replication_connect_timeout = 10,
+--       replication_sync_lag = 0.01,
+--       replication_connect_quorum = 3,
+--       replication = {
+--           Server.build_listen_uri('replica1'),
+--           Server.build_listen_uri('replica2'),
+--           Server.build_listen_uri('replica3'),
+--       },
+--   }
+--   local replica_set = ReplicaSet:new({
+--       servers = {
+--           {alias = 'replica1', box_cfg = box_cfg},
+--           {alias = 'replica2', box_cfg = box_cfg},
+--           {alias = 'replica3', box_cfg = box_cfg},
+--       }
+--   })
+--   replica_set:start()
+--   replica_set:wait_for_fullmesh()
+function ReplicaSet:new(object)
+    if not object then object = {} end
+    self:inherit(object)
+    object:initialize()
+    return object
+end
+
+-- Initialize the replica set object.
+function ReplicaSet:initialize()
+    self._server = Server
+    if self.servers then
+        local configs = table.deepcopy(self.servers)
+        self.servers = {}
+        for _, config in ipairs(configs) do
+            self:build_and_add_server(config)
+        end
+    else
+        self.servers = {}
+    end
+end
+
+--- Build a server object for the replica set.
+--
+-- @tab[opt] config Configuration for the new server.
+-- @return table
+-- @see luatest.server:new
+function ReplicaSet:build_server(config)
+    checks('table', self._server.constructor_checks)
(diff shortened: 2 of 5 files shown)


config.ld
```diff
@@ -6,9 +6,10 @@ file = {
...
     'luatest/runner.lua',
     'luatest/server.lua',
+    'luatest/replica_set.lua',
 }
 topics = {
     'CHANGELOG.md',
 }
 format = 'markdown'
-sort = true
+sort = 'true'
```
NickVolynkin, line 15: If I remember correctly, the sorting above is intentional. Or is it about items within a class?

luatest/server.lua
```diff
@@ -59,7 +59,7 @@ end
 
 --- Build a server object.
 --
--- @tab object
+-- @tab[opt] object
```
NickVolynkin, line 62: The 'object' is not explained. Is it some prototype object to base the new server object on?

luatest/replica_set.lua
```diff
@@ -0,0 +1,231 @@
...
+            for _, server2 in ipairs(replica_set.servers) do
+                if server1 ~= server2 then
+                    local server1_id = server1:get_instance_id()
+                    local server2_id = server2:get_instance_id()
+                    if server1_id ~= server2_id then
+                        server1:assert_follows_upstream(server2_id)
+                    else
+                        -- If IDs are equal, nodes are anonymous replicas and
+                        -- not registered yet. Raise an error to retry checking
+                        -- full mesh again.
```
NickVolynkin, line 222: So if we found anonymous replicas, we start the search again from the start? Could we wait and retry on this particular pair and then move on? Or is it ineffective?

luatest/replica_set.lua
```diff
@@ -0,0 +1,231 @@
...
+    table.insert(self.servers, server)
+end
+
+--- Build a server object and add it to the replica set.
+--
+-- @tab[opt] config Configuration for the new server.
+-- @return table
+-- @see luatest.server:new
+function ReplicaSet:build_and_add_server(config)
+    checks('table', self._server.constructor_checks)
```
NickVolynkin, line 99: this check is a part of the `build_server(object)` method. Why make it just before calling that method?

NickVolynkin: Okay, I thought it was some kind of assert. Is it something else?

luatest/replica_set.lua
```diff
@@ -0,0 +1,231 @@
...
+    checks('table', 'string')
+    for _, server in ipairs(self.servers) do
+        if server.alias == alias then
+            return server
+        end
+    end
+    return nil
+end
+
+local function get_server_index_by_alias(servers, alias)
```
NickVolynkin, line 119: `@return int|nil`?

NickVolynkin: Btw, why is it local? Can't it be useful in tests?

luatest/replica_set.lua
```diff
@@ -0,0 +1,231 @@
...
+--- Delete the server object from the replica set by the server alias.
+--
+-- @string alias Server alias.
+function ReplicaSet:delete_server(alias)
+    checks('table', 'string')
+    local server_index = get_server_index_by_alias(self.servers, alias)
+    if server_index then
+        table.remove(self.servers, server_index)
+    else
+        log.warn(('Server with alias "%s" does not exist in replica set')
```
NickVolynkin, line 137: something to identify the replicaset?

luatest/replica_set.lua
```diff
@@ -0,0 +1,231 @@
...
+function ReplicaSet:inherit(object)
+    setmetatable(object, self)
+    self.__index = self
+    return object
+end
+
+--- Build a replica set object.
+--
+-- @tab[opt] object
+-- @tab[opt] object.servers List of server configurations to build server
```
NickVolynkin, line 22: maybe, `server_configs`? The name `servers` is already used by internal field with a table of `luatest.server` objects.

luatest/replica_set.lua
```diff
@@ -0,0 +1,231 @@
...
+
+    local wait_until_ready = true -- the option is `true` by default
+    if opts ~= nil and opts.wait_until_ready ~= nil then
+        wait_until_ready = opts.wait_until_ready
+    end
+    if wait_until_ready then
+        for _, server in ipairs(self.servers) do
+            server:wait_until_ready()
+        end
+    end
```
NickVolynkin, line 165:
```suggestion
    if opts == nil or opts.wait_until_ready ~= false then
        for _, server in ipairs(self.servers) do
            server:wait_until_ready()
        end
    end
```

|                    | opts == nil | opts ~= nil |
|--------------------|-------------|-------------|
| opts.wait == nil   | wait        | wait        |
| opts.wait == true  | wait        | wait        |
| opts.wait == false | wait        | not wait    |

luatest/replica_set.lua
ylobankov: Probably we could. But I wouldn't like to introduce the new logic for now.

luatest/server.lua

config.ld
ylobankov: It's about items within a class.

luatest/replica_set.lua
NickVolynkin: understood, thanks

luatest/replica_set.lua
```diff
@@ -0,0 +1,227 @@
...
+function ReplicaSet:start(opts)
+    checks('table', {wait_until_ready = '?boolean'})
+
+    for _, server in ipairs(self.servers) do
+        if not server.process then
+            server:start({wait_until_ready = false})
+        end
+    end
+
+    if not opts or opts.wait_until_ready then
```
NickVolynkin, line 157: what if `opts ~= nil` but `opts.wait_until_ready == nil`?

ylobankov: In this case, we will not wait until the server is ready. But looks like we should wait until the server is ready unless `opts.wait_until_ready` is `false`. Fixed.

NickVolynkin (approved):

PR #277, opened 2022-12-23
Autorequire luatest in server instances

ochaplashkin:
Luatest module available without explicit require:

```lua
g.server:exec(function()
   t.assert(...)
end)
```

Resolves #233

- [x] Tests
- [x] Changelog
- [x] Documentation

Issue #233, opened 2022-03-31
Autorequire luatest in all instances to simplify `:exec()` calls


Gerold103:

Totktonada:
@H9660 This task is not assigned to anyone at the moment. You can just propose an implementation in a pull request.

ochaplashkin:
## The process of solving the issue

### 1. Quick solution

Added the variable `t` to the `server_instance.lua` file

```lua
-- server_instance.lua
...
box.cfg(box_cfg())
box.schema.user.grant('guest', 'super', nil, nil, {if_not_exists = true})

t = require('luatest')
...
```
**Error:** this only works if the server is created based on the server_instance.lua. Other (clean, custom) servers are not working.

### 2. Changing the server instance 

```lua
-- server.lua
function Server:exec(fn, args, options)
  ...
  local t = require('luatest')
  return exec_tail(self:eval([[
          local dump, args = ...
          local fn = loadstring(dump)
  ...
```

**Error:** incorrect scope of the `t` variable.

### 3. Declaring a local variable on the server

```lua
-- server.lua
function Server:exec(fn, args, options)
  ...
  return exec_tail(self.net_box:eval([[
          local t = require('luatest')
          local dump, args = ...
          local fn = loadstring(dump)
  ...
```
**Error:** upvalues exception in lua

### 4. Declaring a global variable on the server

```lua
-- server.lua
function Server:exec(fn, args, options)
  ...
  return exec_tail(self.net_box:eval([[
          local t = require('luatest')
          rawset(_G, 't', _t)
          local dump, args = ...
          local fn = loadstring(dump)
  ...
```
**Error:** if the environment variable `LUA_PATH` is not set, then nothing works (and does not say about it)

### 5. Added import via call and logging

```lua
-- server.lua
function Server:exec(fn, args, options)
  ...
  return exec_tail(self.net_box:eval([[
      local is_ok, _t = pcall(require, 'luatest')
      if is_ok then
            rawset(_G, 't', _t)
        else
            require('log').warn('LUA_PATH is unset, ' .. _t)
        end
     local dump, args = ...
     local fn = loadstring(dump)
     ...
     ]]))
end
```

It looks fine. However, we noticed that the `exec` method uses `eval`.

### Final. Move our solution to the `eval` method

```lua
--server.lua
function Server:eval(...)
    ...
    self.net_box:eval([[
        local is_ok, _t = pcall(require, 'luatest')
        if is_ok then
            rawset(_G, 't', _t)
[...]

PR #289, opened 2023-02-02
Improve luatest auto-requiring in `Server:exec`

ylobankov:
In the commit tarantool/luatest@58eae75 auto-requiring of the 'luatest'
lib was added to the `Server:eval()` function, so it was also available
in `Server:exec()`, and it was defined as `t` variable. In the process
of development we found that the following construction doesn't work due
to the upvalue error:

    -- foo_bar_test.lua
    local t = require('luatest')

    ...
    g.test_foo_bar = function()
        server:exec(function()
            t.assert_equals(1, 1)  -- upvalues exception (!)
        end)
    end

To fix this error we decided to make the global variable `t` that had
the 'luatest' lib already loaded. So there was no need to import the
'luatest' lib in the test files and the following example worked:

    -- foo_bar_test.lua
    -- local t = require('luatest')  -- `t` is already defined

    ...
    g.test_foo_bar = function()
        server:exec(function()
            t.assert_equals(1, 1)  -- it works
        end)
    end:

But such an approach is not explicit enough and may mislead developers.
So it was decided to improve and extend the existing mechanism. Now we
just make the needed upvalue with the 'luatest' lib inside available in
the `Server:exec()` function via some magic. And now we can do something
like this:

    -- foo_bar_test.lua
    local t1 = require('luatest')
    local t2 = require('luatest')

    ...
    g.test_foo_bar_1 = function()
        server:exec(function()
            t1.assert_equals(1, 1)  -- it works
        end)
    end

    g.test_foo_bar_2 = function()
        server:exec(function()
            t2.assert_equals(1, 1)  -- it works
        end)
    end

Unfortunately, we cannot do the same for the `Server:eval()` function
and now it becomes just a shortcut for `Server.net_box:eval()` as it
was before the first version of auto-requiring of the 'luatest' lib.

Follows up #277
In scope of #233

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,6 +1,4 @@
 include_files = {"**/*.lua", "*.rockspec", "*.luacheckrc"}
 exclude_files = {"lua_modules/", ".luarocks/", ".rocks/", "tmp/"}
 
-globals = {"t"}
-
 max_line_length = 120
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -472,7 +472,7 @@ end
 
 --- Evaluate Lua code on the server.
 --
--- This is a wrapper for `server.net_box:eval()`.
+-- This is a shortcut for `server.net_box:eval()`.
 -- @string code
 -- @tab[opt] args
 -- @tab[opt] options
@@ -480,17 +480,7 @@ function Server:eval(code, ...)
     if self.net_box == nil then
         error('net_box is not connected', 2)
     end
-    local preamble = [[
-        if not rawget(_G, 't') then
-            local is_ok, _t = pcall(require, 'luatest')
-            if is_ok then
-                rawset(_G, 't', _t)
-            else
-                require('log').warn('LUA_PATH is unset or incorrect, ' .. _t)
-            end
-        end
-    ]]
-    return self.net_box:eval(preamble .. code, ...)
+    return self.net_box:eval(code, ...)
 end
 
 --- Call remote function on the server by name.
@@ -535,8 +525,9 @@ end
 --    end, {1, 2})
 --    -- sum == 3
 --
+--    local t = require('luatest')
 --    server:exec(function()
---        -- luatest is always available as `t`
+--        -- luatest is available via `t` upvalue
 --        t.assert_equals(math.pi, 3)
 --    end)
 --    -- mytest.lua:12: expected: 3, actual: 3.1415926535898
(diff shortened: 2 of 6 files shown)


.luacheckrc
```diff
@@ -1,6 +1,4 @@
 include_files = {"**/*.lua", "*.rockspec", "*.luacheckrc"}
```
Totktonada, line 1: Nit: It seems, we should adjust the wording in the changelog a bit.

luatest/server.lua
```diff
@@ -547,24 +538,44 @@ function Server:exec(fn, args, options)
...
     end
 
-    return exec_tail(self:eval([[
-        local dump, args = ...
+    return exec_tail(self.net_box:eval([[
+        local dump, args, autorequired_ups = ...
         local fn = loadstring(dump)
+        for i, v in pairs(autorequired_ups) do
+            debug.setupvalue(fn, i, require(v))
+        end
```
Totktonada, line 572: I'm not sure that upvalues numeration is guaranteed to be the same for the original function and its `loadstring()` of `dump()`. I would rather loop over existing upvalues and match it with a module name by the upvalue name.

luatest/server.lua
```diff
@@ -547,24 +538,44 @@ function Server:exec(fn, args, options)
...
+        local is_other_up = true
+        local name, value = debug.getupvalue(fn, i)
+        for _, p in ipairs(autorequired_pkgs) do
+            if package.loaded[p] == value then
+                autorequired_ups[i] = p
+                is_other_up = false
+                break
+            end
+        end
+        if is_other_up then
```
Totktonada, line 554: Nit: It can be chacked as `autorequired_ups[i] ~= nil` without extra `is_other_up` variable. (Or as `autorequired_ups[name] ~= nil` if you'll accept my comment regarding upvalues matching by names.)

.luacheckrc
ylobankov: Fixed.

luatest/server.lua

luatest/server.lua

.luacheckrc
Totktonada: Shouldn't we say about `:exec()` call, not just 'server instance'?

luatest/server.lua
```diff
@@ -547,24 +538,49 @@ function Server:exec(fn, args, options)
...
+        local nups = debug.getinfo(fn, 'u').nups
+        for up_name, pkg_name in pairs(passthrough_ups) do
+            for i = 1, nups do
+                local name, value = debug.getupvalue(fn, i)
+                if name == up_name then
+                    debug.setupvalue(fn, i, require(pkg_name))
+                    break
+                end
+            end
+        end
```
Totktonada, line 577: If I don't miss anything, we can look over upvalues and lookup the `passthrough_ups` table by the upvalue name (without the nested loop).

.luacheckrc
ylobankov: Fixed.

Totktonada (approved):

PR #294, opened 2023-03-24
Add assert_one_of

yngvar-antonsson:
Add `assert_one_of` to assert that the value is one of several alternatives.


DifferentialOrange (approved):
I don't quite get why I was summoned to review luatest PR since I never contributed here and don't know how the infrastructure works, but the patch seems fine.

You can also think about unifying the code with `assert_covers` since they do similar things.

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@
 - Raise an error in the `Server:wait_for_condition()` function when
   the server process is terminated. This is useful to not wait for timeout, for example,
   when a server fails to start due to bad configuration.
+- Add `assert_one_of`.
```
ylobankov, line 53: Could you please change it to "Add new assert function `assert_one_of`"?

README.rst
```diff
@@ -170,6 +170,8 @@ List of luatest functions
 +--------------------------------------------------------------------+-----------------------------------------------+
 | ``assert_covers (actual, expected[, message])``                    | Checks that actual map includes expected one. |
 +--------------------------------------------------------------------+-----------------------------------------------+
+| ``assert_one_of (actual, expected[, message])``                    | Checks that value is of on alternatives.      |
```
ylobankov, line 173: Shouldn't `is of on alternatives` be `is one of alternatives`?

luatest/assertions.lua
```diff
@@ -276,6 +276,25 @@ function M.assert_ge(left, right, message)
     end
 end
 
+--- Check that value is one of alternates.
```
ylobankov, line 279: In README you use `alternatives`. Let's be consistent.

Totktonada:
AFAIK we generaly follow the LuaUnit assertion naming (but just snake case variants). There are `assert_table_contains` and `assert_table_not_contains` ([code](https://github.com/bluebird75/luaunit/blob/6f526f85faabdca7cad893ff4d5f61e4af5d0dd5/luaunit.lua#L1940C36-L1941])

DifferentialOrange:
> AFAIK we generaly follow the LuaUnit assertion naming (but just snake case variants). There are `assert_table_contains` and `assert_not_table_contains` ([code](https://github.com/bluebird75/luaunit/blob/6f526f85faabdca7cad893ff4d5f61e4af5d0dd5/luaunit.lua#L1940C36-L1941%5D)).

For me `assert_table_contains` and `assert_one_of` have different ideas, even though their implementation is the same.

Totktonada:
BTW, the implementation is not exactly same. You're using `ipairs()`, so the function is applicable only for array alike tables. However, LuaUnit uses `pairs()` and so it makes the function applicable for tables like `{foo = 'needle', bar = 'unrelated'}`.

Regarding `assert_one_of` name semantic: if the idea is to check that the given element is present in the given table exactly one time, the implementation doesn't check it.

However, I still think that LuaUnit names are good enough and I see no reason to use different ones.

PR #303, opened 2023-04-13
Add error handling at the server instance level

### LINKED ISSUE #242: Wrong error returned by `luatest.Server.exec` if transaction was active
If an error is thrown by the executed function while transaction is active, `luatest.Server.exec` returns a wrong error: "Transaction is active at return from function".

To reproduce the issue, add the following test to the [Tarantool](https://github.com/tarantool/tarantool) repository and run it:

```Lua
local server = require('test.luatest_helpers.server')
local t = require('luatest')
local g = t.group()

g.before_all = function(cg)
    cg.server = server:new({alias = 'master'})
    cg.server:start()
end

g.after_all = function(cg)
    cg.server:stop()
end

g.test_error_in_tx_net_box = function(cg)
    local c = require('net.box').connect(cg.server.net_box_uri)
    t.assert_error_msg_contains(
        "My error", c.eval, c,
        [[box.begin() error("My error")]])
end

g.test_error_in_tx_eval = function(cg)
    t.assert_error_msg_contains(
        "My error", cg.server.eval, cg.server,
        [[box.begin() error("My error")]])
end

g.test_error_in_tx_exec = function(cg)
    t.assert_error_msg_contains(
        "My error", cg.server.exec, cg.server,
        function() box.begin() error("My error") end)
end
```

You'll get:

```
$ luatest -v test/box-luatest/error_in_tx_test.lua
Running with --shuffle group:7246
Started on Fri Aug 26 11:10:58 2022
    box-luatest.error_in_tx.test_error_in_tx_exec ... (0.001s) fail
...arantool/tarantool/test/box-luatest/error_in_tx_test.lua:28: Error message does not contain: "My error"
Error message received: "Transaction is active at return from function"

    box-luatest.error_in_tx.test_error_in_tx_net_box ... (0.005s) Ok
    box-luatest.error_in_tx.test_error_in_tx_eval ... (0.001s) Ok
=========================================================

Failed tests:
-------------

1) box-luatest.error_in_tx.test_error_in_tx_exec
...arantool/tarantool/test/box-luatest/error_in_tx_test.lua:28: Error message does not contain: "My error"
Error message received: "Transaction is active at return from function"

stack traceback:
        ...arantool/tarantool/test/box-luatest/error_in_tx_test.lua:28: in function 'box-luatest.error_in_tx.test_error_in_tx_exec'
        ...
        [C]: in function 'xpcall'

Ran 3 tests in 0.232 seconds, 2 succeeded, 1 failed

=========================================================
Failed tests:

box-luatest.error_in_tx.test_error_in_tx_exec
```

Note, neither `net.box.eval` nor `luatest.Server.eval` is affected (probably fixed in https://github.com/tarantool/tarantool/issues/7288).

ochaplashkin:
When executing the following code on the remote server instance using `server:exec()`, we could get an unobvious error:
```lua
s = Server:new()
s:start()

-- we expected a custom error with the "My error" message, but
-- got "Transaction is active at return from function"
s:exec(function() box.begin() error("My error") end)
```

This is due to the specific evaluation and capture of the return values of `net_box:eval` (more detailed see here: https://github.com/tarantool/tarantool/issues/7288).

The error will be raised on the server instance side. In the process of development we found that our approach does not work in the specific case:

```lua
-- an error message has been expected but got an error table:
-- {
--	class = LuatestError,
--	status = fail,
--	message = "a value evaluating to true, actual: false"
-- }
s:exec(function() require('luatest').assert(false) end)
```
To fix this error we decided to add the [encoding an error to JSON format on the server](https://github.com/tarantool/luatest/pull/303/files#diff-675c7efebade428c1f2415794641c31a928f06bd76226cd648f85788137af85bR632) and decode it [on the client side](https://github.com/tarantool/luatest/pull/303/files#diff-675c7efebade428c1f2415794641c31a928f06bd76226cd648f85788137af85bR525).

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #242

### DIFF
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -522,7 +522,8 @@ end
 
 local function exec_tail(ok, ...)
     if not ok then
-        error(..., 0)
+        local _ok, res = pcall(json.decode, tostring(...))
+        error(_ok and res or ..., 0)
     else
         return ...
     end
@@ -604,7 +605,14 @@ function Server:exec(fn, args, options)
             :format(utils.get_fn_location(fn)))
     end
 
-    return exec_tail(self.net_box:eval([[
+    -- The function `fn` can return multiple values and we cannot use the
+    -- classical approach to work with the `pcall`:
+    --
+    --    local status, result = pcall(function() return 1, 2, 3 end)
+    --
+    -- `result` variable will contain only `1` value, not `1, 2, 3`.
+    -- To solve this, we put everything from `pcall` in a table.
+    return exec_tail(pcall(self.net_box.eval, self.net_box, [[
         local dump, args, passthrough_ups = ...
         local fn = loadstring(dump)
         for i = 1, debug.getinfo(fn, 'u').nups do
@@ -613,11 +621,19 @@ function Server:exec(fn, args, options)
                 debug.setupvalue(fn, i, require(passthrough_ups[name]))
             end
         end
+        local result
         if args == nil then
-            return pcall(fn)
+            result = {pcall(fn)}
         else
-            return pcall(fn, unpack(args))
+            result = {pcall(fn, unpack(args))}
         end
+        if not result[1] then
+            if type(result[2]) == 'table' then
+                result[2] = require('json').encode(result[2])
+            end
+            error(result[2], 0)
+        end
+        return unpack(result, 2)
     ]], {string.dump(fn), args, passthrough_ups}, options))
 end
 
--- a/test/server_test.lua
+++ b/test/server_test.lua
@@ -429,3 +429,50 @@ end
 g.after_test('test_netbox_uri_is_not_overridden', function()
     g.s1:drop()
 end)
+
+g.before_test('test_error_level_is_correct', function()
+    g.s = Server:new()
+    g.s:start()
+end)
+
+g.test_error_level_is_correct = function()
+    local c = require('net.box').connect(g.s.net_box_uri)
+
+    t.assert_error_msg_contains( -- error in exec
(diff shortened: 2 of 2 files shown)


ylobankov (approved):
LGTM

Totktonada:
It was not qute obvious for me what is going on here. I talked with Yaroslav and I understand it now. The summary below is for ones who stuck here like I am.

## Terms

* **Server's code** is the part of `<server>:exec()` call that is evaluated on the remote server.
* **User's code** is the part of server's code, which is provided by a user as a `<server>:exec()` parameter.
* **Client's code** is the part of `<server>:exec()` that is executed on the client (usually in the main code of the test) -- before and after the `<net_box>:eval()` call.

## Old behavior

In the old server's code executes user's code under pcall and returns pcall's result. The client's code re-raises an error if needed.

If an active transaction is left by the user's code (`box.begin()` without `box.commit()` or `box.rollback()`), tarantool acts like the `Transaction is active at return from function` error was raised from the function.

As result, due to `pcall`, we got the error regarding the active transaction instead of our one.

## New behavior

It seems like everything that we should do is re-raising errors from pcall in the server's code. It would solve the original problem.

However another problem appears: a table error like `error({<...>})` becomes a string like `'table: 0x407927c0'` on the client. Taking into account that luatest's asserts raise table errors, it hits many user scenarious.

So we need to do a trick: encode the table error into JSON on the server (and raise this JSON string as an error), pcall it on the client, decode from JSON and re-raise.

That's why all this strange code is here.

Issue #306, opened 2023-05-12
--shuffle default is not 'none'


Totktonada:
However, the `--help` output states that it is `none`.

See the initial investigation here: https://github.com/tarantool/test-run/pull/380#issuecomment-1545507406.

PR #307, opened 2023-05-15
Not require own test helper when running tests

ylobankov:
> Fow what test/helper.lua is needed?

Luatest automatically requires `test/helper.lua` file if it's present.
Usually, it is used to configure luatest or run any bootstrap code.

> What's the problem?

To run tarantool functional tests we use the test-run framework [1].
test-run has an integration with luatest (uses luatest as a submodule)
and allows us to run tarantool functional tests written in the luatest
approach. Most of the such tests can be run via luatest directly, but
test-run is used to run all types of tarantool tests at the same time
(diff, python, app, luatest).

When test-run works, it tunes some environment variables for own needs
and sets LUA_PATH to something like this:

    LUA_PATH='<...>/tarantool/?/init.lua;<...>/tarantool/?.lua;\
        <...>/tarantool/test-run/lib/luatest/?/init.lua;\
        <...>/tarantool/test-run/lib/luatest/?.lua;;'

The tarantool tests don't have own `test/helper.lua` file and this file
is required from the luatest submodule of test-run in accordance with
the configured LUA_PATH variable, which is obviously wrong.

> What's the solution?

The simplest way to resolve this issue is to move `test/helper.lua` to
`test/helpers/general.lua` to not auto-require it while running tests.

Actually, there is no reason to have `test/helper.lua` in luatest. I can
assume it was added to run unit tests with `shuffle = 'group'`. But we
can run unit tests with the `--shuffle group` option and have the same
effect.

[1] https://github.com/tarantool/test-run

Closes https://github.com/tarantool/luatest/issues/306

### DIFF
--- a/test/assertions_test.lua
+++ b/test/assertions_test.lua
@@ -1,7 +1,7 @@
 local t = require('luatest')
 local g = t.group()
 
-local helper = require('test.helper')
+local helper = require('test.helpers.general')
 
 g.test_custom_errors = function()
     local function assert_no_exception(fn)
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -33,4 +33,4 @@ jobs:
         run: make lint
 
       - name: Run tests
-        run: bin/luatest -v
+        run: bin/luatest -v --shuffle group
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ add_custom_target(doc
 )
 
 add_custom_target(lint DEPENDS bootstrap COMMAND .rocks/bin/luacheck .)
-add_custom_target(selftest DEPENDS bootstrap COMMAND bin/luatest)
+add_custom_target(selftest DEPENDS bootstrap COMMAND bin/luatest --shuffle group)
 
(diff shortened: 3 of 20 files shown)


test/assertions_test.lua
```diff
@@ -1,7 +1,7 @@
 local t = require('luatest')
 local g = t.group()
 
-local helper = require('test.helper')
+local helper = require('test.helpers.helper')
```
Totktonada, line 4: Nit: helpers.helper is a bit repetitive. Especially because it is the only helper :)

test/helpers/helper.lua
```diff
@@ -2,8 +2,6 @@ local t = require('luatest')
 local Runner = require('luatest.runner')
 local utils = require('luatest.utils')
 
-t.configure({shuffle = 'group'})
```
Totktonada, line 5: This change is not commented anywhere.

Totktonada (approved):
I have no deep understanding of this code, but I don't mind.

test/helpers/helper.lua
ylobankov: Fixed.

ylobankov: Extended commit message.

Issue #124, opened 2021-03-29
Rewrite Process in Popen


olegrok:
Also such approach should preserve backward compatibility (old versions should use old code). But new Tarantool versions 2.4 should works fine. Maybe it will fix https://github.com/tarantool/luatest/issues/120 .

ochaplashkin:
During solution #312 , we tried to rewrite the module, but decided to postpone it indefinitely.
These are notes about what we managed to find out (for the future):

- list of possible problems with [popen](https://github.com/tarantool/tarantool/issues?q=is%3Aopen+is%3Aissue+label%3Apopen);
- `ffi pipe` and `popen popen` are not compatible (by interfaces);
- `output_beatifier:hijack_output(...)` not needed because `POPEN` will do it;
   https://github.com/tarantool/luatest/blob/89da427f8bb3bb66e01d2a7b5a9370d0428d8c52/luatest/process.lua#L64-L66
- processing of arguments and environment variables is not necessary:
   https://github.com/tarantool/luatest/blob/89da427f8bb3bb66e01d2a7b5a9370d0428d8c52/luatest/process.lua#L67-L68
- there are some limitations in `popen:new()`, you need to use `popen:shell()`;
- you can replace the `read` function (a more correct option):
https://github.com/tarantool/luatest/blob/89da427f8bb3bb66e01d2a7b5a9370d0428d8c52/luatest/ffi_io.lua#L32-L47
- see #312 about this logic (diagram)
   https://github.com/tarantool/luatest/blob/89da427f8bb3bb66e01d2a7b5a9370d0428d8c52/luatest/process.lua#L51-L59
- `ffi.gc` works correctly with `popen`
   https://github.com/tarantool/luatest/blob/89da427f8bb3bb66e01d2a7b5a9370d0428d8c52/luatest/process.lua#L78-L81

olegrok:
> ffi pipe and popen popen are not compatible (by interfaces);

This issue was written when 1.10 was relevant. Currently it's not so. You don't need to care about ffi and popen compatibility.

PR #312, opened 2023-06-08
Fix detect tarantool crash at exit

ochaplashkin:
### What's the problem?

During the shutdown of the server, the server process is polled. If the call `ffi.C.kill(<pid>, 0)` returns the value 0, then the process will be considered alive, otherwise it will be dead.

Luatest cannot distinguish the states when the process is completed correctly and incorrectly. This means that if a critical error appears at the end of the process (e.g., segmentation fault), then stopping the server will be considered successful (and the test-case too).

See sequence diagram for the more details:
> FYI: `par` section - parallel activity

luatest_start_stop_crash

### What's the solution?

Possible solutions:

1. rewrite the process.lua module to the popen library (not in our case, see comments for more details tarantool/luatest#124);
2. **check the stderr of the process when the server is stopped.**

The second option was chosen.

Each process contains its own instance of output_beautifier (which reads **stdout** and **stderr**). Now this instance will contain the stderr field (string type) where the entire **stderr** from the process will be saved.

During the server shutdown, it looks for the substring `segmentation fault` and return an error if the substring is found:

```
Failure in after_all hook: ...ace/tarantool/test-run/lib/luatest/luatest/server.lua:371: Segmentation fault during process termination (alias: master, workdir: master-SrQjEIiLyRHS, pid: 152323)
Segmentation fault
  code: SEGV_MAPERR
  addr: (nil)
  context: 0x7feac9f7f800
  siginfo: 0x7feac9f7f930
  rax      0x0                0
  rbx      0x5600ebb0024d     94561954169421
  rcx      0x5600ebb4e703     94561954490115
  rdx      0x5600eb8708fb     94561951484155
  rsi      0x0                0
  rdi      0x0                0
  rsp      0x7feac9f7fec0     140646387547840
  rbp      0x7feac9f7fed0     140646387547856
  r8       0x7e000            516096
  r9       0x7feae5619280     140646847451776
  r10      0x0                0
  r11      0x246              582
  r12      0x0                0
  r13      0x0                0
  r14      0x0                0
  r15      0x0                0
  rip      0x5600eb870910     94561951484176
  eflags   0x10206            66054
  cs       0x33               51
  gs       0x0                0
  fs       0x0                0
  cr2      0x0                0
  err      0x6                6
  oldmask  0x0                0
  trapno   0xe                14
Current time: 1688036858
Please file a bug at https://github.com/tarantool/tarantool/issues
Attempting backtrace... Note: since the server has already crashed, 
this may fail as well
#1  0x5600ebaf3ee2 in crash_collect+256
#2  0x5600ebaf491e in crash_signal_cb+100
#3  0x7feae5442520 in __sigaction+80
#4  0x5600eb870910 in iproto_on_shutdown_f(void*)+21
#5  0x5600ebb4e73c in trigger_commom_f+57
#6  0x5600ebb28844 in trigger_fiber_f(__va_list_tag*)+188
#7  0x5600eb85fc05 in fiber_cxx_invoke(int (*)(__va_list_tag*), __va_list_tag*)+34
#8  0x5600ebb00328 in fiber_loop+219
#9  0x5600ebe6c25a in coro_init+120

stack traceback:
```

Closes https://github.com/tarantool/luatest/issues/252

### DIFF
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -367,6 +367,17 @@ function Server:stop()
         if not ok and not err:find('Process is terminated when waiting for') then
             error(err)
         end
+        if self.process.output_beautifier.stderr:find('Segmentation fault') then
+            error(
+                ('Segmentation fault during process termination (alias: %s, workdir: %s, pid: %d)\n%s')
+                :format(
+                    self.alias,
+                    fio.basename(self.workdir),
+                    self.process.pid,
+                    self.process.output_beautifier.stderr
+                )
+            )
+        end
         log.debug('Killed server process PID ' .. self.process.pid)
         self.process = nil
     end
--- a/luatest/output_beautifier.lua
+++ b/luatest/output_beautifier.lua
@@ -70,6 +70,7 @@ function OutputBeautifier.mt:initialize()
         self.class.COLOR_BY_NAME[self.color] or
         OutputBeautifier:next_color_code()
     self.pipes = {stdout = ffi_io.create_pipe(), stderr = ffi_io.create_pipe()}
+    self.stderr = ''
 end
 
 -- Replace standard output descriptors with pipes.
@@ -86,7 +87,7 @@ function OutputBeautifier.mt:enable(options)
     end
     self.fibers = {}
     for i, pipe in pairs(self.pipes) do
-        self.fibers[i] = fiber.new(self.run, self, pipe[0])
+        self.fibers[i] = fiber.new(self.run, self, pipe[0], i)
     end
     self.fibers.pid_tracker = options and options.track_pid and fiber.new(function()
         Process = Process or require('luatest.process')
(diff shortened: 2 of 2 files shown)


luatest/server.lua
```diff
@@ -334,7 +334,13 @@ local function wait_for_condition(cond_desc, server, func, ...)
     local deadline = clock.time() + WAIT_TIMEOUT
     while true do
         if not server.process:is_alive() then
-            error(('Process is terminated when waiting for "%s" condition for server (alias: %s, workdir: %s, pid: %d)')
+            -- The error does not appear at our stack level because something sends
+            -- signum with 14 status code and the external signal handler is called.
```
igormunkin, line 338: Well, while digging tarantool/tarantool#8748 we didn't receive `SIGALRM` with @Buristan, so this line is little related to the actual case. I'd rather mentioned two facts:
1. The resulting string length/hash matters (at least, its change provides different results)
2. It's likely related to the recent changes in LuaJIT unwinder scope (I suggest to try your case either with Tarantool 3.0 before tarantool/tarantool@cde911d or with the recent 2.11 release).

Otherwise, LGTM.

ochaplashkin: Now it is clear to us that this is not a _Lua interpreter_ problem, but a defect in the process completion logic.

See [comment](https://github.com/tarantool/tarantool/issues/8748#issuecomment-1613320855).

ylobankov:
Please add a changelog entry.

ylobankov (approved):

ylobankov (approved):

PR #323, opened 2023-09-05
Consider login and password in net box URI

ylobankov:


luatest/server.lua
```diff
@@ -265,6 +243,28 @@ function Server:start(opts)
...
+        end
+        self.datadir = nil
+    end
+    local parsed_net_box_uri = uri.parse(self.net_box_uri)
+    if parsed_net_box_uri.host == 'unix/' then
+        -- Linux uses max 108 bytes for Unix domain socket paths, which means a
+        -- 107 characters string ended by a null terminator. Other systems use
+        -- 104 bytes and 103 characters strings.
+        local max_unix_socket_path = {linux = 107, other = 103}
+        local system = os.execute('[ $(uname) = Linux ]') == 0 and 'linux' or 'other'
```
Totktonada, line 261: While we're here: isn't just `jit.os == 'Linux'` works?

luatest/server.lua
```diff
@@ -265,6 +243,28 @@ function Server:start(opts)
 
     self:initialize()
 
+    fio.mktree(self.workdir)
```
Totktonada, line 247: I would like to see directories creation as a separate method to be able to rewrite it in a child class.

Note: It'll require to separate it from the socket path length error.

luatest/server.lua
```diff
@@ -141,16 +141,6 @@ function Server:initialize()
 
     if self.workdir == nil then
         self.workdir = fio.pathjoin(self.vardir, self.id)
-        fio.rmtree(self.workdir)
```
Totktonada, line 144: I would like to see a motivation of removing of the `rmtree` call.

luatest/server.lua
Totktonada: BTW, the socket path length check can be performed at initialization. It is just validation of the given parameters for correctness.

Totktonada (approved):
No objections from my side (after a brief glance).

luatest/server.lua
ylobankov: This call will be removed in the scope of a separate PR.

luatest/server.lua
ylobankov: I guess it works, but it is not related to my changes :)

luatest/server.lua
ylobankov: > BTW, the socket path length check can be performed at initialization. It is just validation of the given parameters for correctness.

Yes, but in this case we need to parse URI two times. Well, I think it is OK.

ylobankov: > I would like to see directories creation as a separate method to be able to rewrite it in a child class.

Fixed.

luatest/server.lua
Totktonada: And it is faster :)

OK, let's postpone.

Totktonada (approved):

ochaplashkin (approved):
LGTM :rocket:

PR #330, opened 2023-09-26
Fix save artifacts

ochaplashkin:
Since commit [^1] the `Server:save_artifacts()` function was called more than once. Due to this fact we could overwrite already saved artifacts. Added a flag that will ensure that the saving will be executed only once.

There was also a problem when copying artifacts was not performed but the path to the artifacts was formed as a directory with artifacts:

```
    artifacts:
        server -> /tmp/t/artifacts/server-XXX
```

And if we tried to look at these artifacts, we could see this:

```
    $ ls -la /tmp/t/artifacts/server-XXX
    ls: cannot access '/tmp/t/artifacts/server-XXX': No such file or
    directory
```

To show explicitly that the saving failed with an error, the following
string will now be written to the artifacts:

```
    artifacts:
        server -> Failed to copy artifacts for server (alias: server-XXX
        workdir: /tmp/t/artifacts/server-XXX)
```

[^1]: https://github.com/tarantool/luatest/commit/251b35f

### DIFF
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -372,13 +372,20 @@ function Server:restart(params, opts)
 end
 
 -- Save server artifacts by copying the working directory.
--- Throws an error when the copying is not successful.
+-- The save logic will only work once to avoid overwriting the artifacts directory.
+-- If an error occurred, then the server artifacts path will be replaced by the
+-- following string: `Failed to copy artifacts for server (alias: <alias>, workdir: <workdir>)`.
 function Server:save_artifacts()
+    if self.artifacts_saved then
+        return
+    end
     local ok, err = fio.copytree(self.workdir, self.artifacts)
     if not ok then
-        log.error(('Failed to copy artifacts for server (alias: %s, workdir: %s): %s')
-            :format(self.alias, fio.basename(self.workdir), err))
+        self.artifacts = ('Failed to copy artifacts for server (alias: %s, workdir: %s)')
+            :format(self.alias, fio.basename(self.workdir))
+        log.error(('%s: %s'):format(self.artifacts, err))
     end
+    self.artifacts_saved = true
 end
 
 -- Wait until the given condition is `true` (anything except `false` and `nil`).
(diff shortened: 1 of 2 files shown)


luatest/server.lua
```diff
@@ -371,14 +371,21 @@ function Server:restart(params, opts)
...
 end
 
--- Save server artifacts by copying the working directory.
--- Throws an error when the copying is not successful.
+--- Save server artifacts by copying the working directory.
+-- The save logic will only work once to avoid overwriting the artifacts directory.
+-- If an error occurred, then server artifacts path will be replaced follow string:
+-- `Failed to copy artifacts for server (alias: <alias>, workdir: <workdir>)`.
 function Server:save_artifacts()
+    if self.is_artifacts_handled then
```
ylobankov, line 379: What about `artifacts_saved`?

luatest/server.lua
```diff
@@ -371,14 +371,21 @@ function Server:restart(params, opts)
...
+        return
+    end
     local ok, err = fio.copytree(self.workdir, self.artifacts)
     if not ok then
-        log.error(('Failed to copy artifacts for server (alias: %s, workdir: %s): %s')
-            :format(self.alias, fio.basename(self.workdir), err))
+        self.artifacts = ('Failed to copy artifacts for server (alias: %s, workdir: %s)')
+                        :format(self.alias, fio.basename(self.workdir))
```
ylobankov, line 385: Formatting is strange. It should be

```
self.artifacts = ('Failed to copy artifacts for server (alias: %s, workdir: %s)')
    :format(self.alias, fio.basename(self.workdir))
```

luatest/server.lua
```diff
@@ -371,14 +371,21 @@ function Server:restart(params, opts)
     log.debug('Restarted server PID: ' .. self.process.pid)
 end
 
--- Save server artifacts by copying the working directory.
--- Throws an error when the copying is not successful.
+--- Save server artifacts by copying the working directory.
+-- The save logic will only work once to avoid overwriting the artifacts directory.
+-- If an error occurred, then server artifacts path will be replaced follow string:
```
ylobankov, line 376: `then the server artifacts path will be replaced by the following string`

ylobankov (approved):

luatest/server.lua
ochaplashkin: Fixed.

luatest/server.lua
ochaplashkin: Fixed.

luatest/server.lua
ochaplashkin: Fixed.

PR #338, opened 2023-11-20
Fix running unit tests with coverage enabled

ylobankov:
Before this patch, it was impossible to run luatest unit tests with the coverage enabled because some tests failed with the following error:

    luatest/test/server_test.lua:393: net_box is not connected
    stack traceback:
        ...
        [C]: in function 'xpcall'
    artifacts:
        server -> /tmp/t/artifacts/server-KLiHQDTbAJ4b
        server -> /tmp/t/artifacts/server-QQQzEqDr4Hsy

This patch fixes the issue.

Also, it updates the name of the `make` target for running unit tests with coverage to `selftest-coverage` and enables coverage testing on CI.

### DIFF
--- a/.github/workflows/test_on_push.yaml
+++ b/.github/workflows/test_on_push.yaml
@@ -32,5 +32,5 @@ jobs:
       - name: Run linter
         run: make -C build lint
 
-      - name: Run tests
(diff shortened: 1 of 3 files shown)


ochaplashkin (approved):
LGTM

Issue #341, opened 2023-12-11
URI length is checked instead of unix socket path length


ImeevMA:
For example we have test `app-luatest/example_test.lua` on Tarantool-Enterprise:
```Lua
local server = require('luatest.server')
local t = require('luatest')

local g = t.group()

g.test_too_long_uri = function()
    local net_box_uri = "guest@unix/:/tmp/temp?transport=ssl&" ..
                        "ssl_cert_file=/home/mergen/work/ee/test/enterprise-luatest/ssl_cert/client.crt&" ..
                        "ssl_key_file=/home/mergen/work/ee/test/enterprise-luatest/ssl_cert/client.key"
    t.assert_equals(#net_box_uri, 192)
    g.server = server:new{
        alias = 'master',
        net_box_uri = net_box_uri,
    }
    g.server:start()
    g.server:drop()
end
```

When we run it, we will get an error:
```
======================================================================================
WORKR TEST                                            PARAMS          RESULT
---------------------------------------------------------------------------------
[001] app-luatest/example_test.lua                                    [ fail ]
[001] Test failed! Output from reject file /tmp/t/rejects/app-luatest/example.reject:
[001] Tarantool version is 3.0.0-beta1-53-g59b817ef8
[001] TAP version 13
[001] 1..1
[001] # Started on Tue Dec 12 14:40:48 2023
[001] # Starting group: app-luatest.example
[001] not ok 1	app-luatest.example.test_too_long_uri
[001] #   .../mergen/work/dev/test-run/lib/luatest/luatest/server.lua:165: Net box URI must be <= max Unix domain socket path length (107 chars)
[001] #   stack traceback:
[001] #   	/home/mergen/work/dev/test/app-luatest/example_test.lua:11: in function 'app-luatest.example.test_too_long_uri'
[001] #   	...
[001] #   	[C]: in function 'xpcall'
[001] # Ran 1 tests in 0.003 seconds, 0 succeeded, 1 errored
[001] 
[001] [test-run server "luatest_server"] The log file /tmp/t/001_app-luatest/example_test.log has zero size
[Main process] Got failed test; gently terminate all workers...
[001] Worker "001_app-luatest" got failed test; stopping the server...
---------------------------------------------------------------------------------
```

The Unix socket path here is `/tmp/temp` and its length is 9.

ImeevMA:
Fixing this issue will avoid updating box.cfg.listen after server startup if we are using SSL, and instead start the server with the correct box.cfg.listen with SSL.

PR #344, opened 2023-12-12
server: change error message for unix socket path

ImeevMA:
This patch changes the error message when the Unix socket path length exceeds a certain limit.

Follow-up #341

### DIFF
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -164,8 +164,9 @@ function Server:initialize()
         local max_unix_socket_path = {linux = 107, other = 103}
         local system = os.execute('[ $(uname) = Linux ]') == 0 and 'linux' or 'other'
         if parsed_net_box_uri.unix:len() > max_unix_socket_path[system] then
-            error(('Net box URI must be <= max Unix domain socket path length (%d chars)')
-                :format(max_unix_socket_path[system]))
+            error(('Unix domain socket path cannot be longer than %d chars. ' ..
+                   'Current path is: %s'):format(max_unix_socket_path[system],
+                                                 parsed_net_box_uri.unix))
         end
     end
     if type(self.net_box_uri) == 'table' then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Added `assert_error_covers`.
 - Add more logs (gh-326).
 - Add `justrun` helper as a tarantool runner and output catcher (gh-365).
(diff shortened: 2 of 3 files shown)


luatest/server.lua
```diff
@@ -163,8 +163,8 @@ function Server:initialize()
         local max_unix_socket_path = {linux = 107, other = 103}
         local system = os.execute('[ $(uname) = Linux ]') == 0 and 'linux' or 'other'
         if parsed_net_box_uri.unix:len() > max_unix_socket_path[system] then
-            error(('Net box URI must be <= max Unix domain socket path length (%d chars)')
-                :format(max_unix_socket_path[system]))
+            error(('Unix domain socket path length must be not greater than ' ..
+                   '%d chars'):format(max_unix_socket_path[system]))
```
Totktonada, line 167: While we're here, I would add the the problematic path into the message to make it easier to find a place in the code that triggers the problem.

ylobankov: What is the reason for changing the error message?

ylobankov (approved):

luatest/server.lua
ImeevMA: Old message was about whole URI, but only Unix socket path have the limitation.

ylobankov: Net box URI may contain creds, got it, thanks.

ImeevMA: Fixed.

Totktonada (approved):
No objections from my side.

luatest/server.lua
```diff
@@ -163,8 +163,9 @@ function Server:initialize()
         local max_unix_socket_path = {linux = 107, other = 103}
         local system = os.execute('[ $(uname) = Linux ]') == 0 and 'linux' or 'other'
         if parsed_net_box_uri.unix:len() > max_unix_socket_path[system] then
-            error(('Net box URI must be <= max Unix domain socket path length (%d chars)')
-                :format(max_unix_socket_path[system]))
+            error(('Unix domain socket path cannot be longer than %d chars. ' ..
+                   'Current path is: "%s"'):format(max_unix_socket_path[system],
```
ylobankov, line 167: I am not sure that the current path should be quoted.

ylobankov:
This change should be reflected in the changelog file.

luatest/server.lua
ImeevMA: Thanks, fixed.

PR #351, opened 2024-01-19
server: restore exec() sparse output support

### LINKED ISSUE #350: Second value after nil is not returned from server.exec
Second value after nil is not returned from `server.exec`. Caused by https://github.com/tarantool/luatest/commit/e6a2093.

```lua
local server = require('luatest.server')
local t = require('luatest')
local g = t.group()

g.before_all = function(lg)
    lg.master = server:new({alias = 'master'})
    lg.master:start()
end

g.after_all = function(lg)
    lg.master:drop()
end

g.test_check = function(lg)
    local a, b = lg.master:exec(function()
        return nil, {message = "msg"}
    end)

    t.assert_equals(a, nil)
    t.assert_equals(b.message, "msg")
end

```

```
[002] replication-luatest/new_test.lua                                [ fail ]
[002] Test failed! Output from reject file /tmp/t/rejects/replication-luatest/new.reject:
[002] Tarantool version is 3.1.0-entrypoint-26-g57b902887
[002] TAP version 13
[002] 1..1
[002] # Started on Tue Jan 16 16:28:19 2024
[002] # Starting group: replication-luatest.new
[002] not ok 1	replication-luatest.new.test_check
[002] #   ...ming/tnt/tarantool/test/replication-luatest/new_test.lua:20: attempt to index local 'b' (a nil value)
[002] #   stack traceback:
[002] #   	...
[002] #   	[C]: in function 'xpcall'
[002] #   artifacts:
[002] #   	master -> /tmp/t/002_replication-luatest/artifacts/master-tQXcu8z2MNYt
[002] # Ran 1 tests in 0.314 seconds, 0 succeeded, 1 errored
[002] 

```

DifferentialOrange:
After [1], returning values after nil (for example, classic `return nil, err` scenario) is broken. This patch restores the behavior.  Since server doesn't always has access to `require('luatest.utils')`, utility is provided through its dump. Since `net.box` already handles `nil` args input with `box.NULL`, everything is fine for exec arguments.

1. https://github.com/tarantool/luatest/commit/e6a2093

I didn't forget about

- [x] Tests
- [x] Changelog
- Documentation (bugfix)

Close #350

### DIFF
--- a/luatest/utils.lua
+++ b/luatest/utils.lua
@@ -191,4 +191,26 @@ function utils.is_tarantool_binary(path)
     return path:find('^.*/tarantool[^/]*$') ~= nil
 end
 
+function utils.unpack_sparse_array(array, index)
+    -- Embed everything so no upvalues required on server.
+    index = index or 1
+
+    local len = 0
+    for k, _ in pairs(array) do
+        if type(k) == 'number' then
+            len = math.max(len, k)
+        end
+    end
+
+    local function unpack_sparse_array_tail(array, index, len) -- luacheck: ignore
+        if index > len then
+            return
+        end
+
+        return array[index], unpack_sparse_array_tail(array, index + 1, len)
+    end
+
+    return unpack_sparse_array_tail(array, index, len)
+end
+
 return utils
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
   * Fixed incorrent unix socket path length check (gh-341).
   * Now net_box_uri can be accepted as table (gh-342).
+  * Fixed returning values from `Server:exec()` if some of them are nil (gh-350).
 
 ## 1.0.0
 
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -686,6 +686,10 @@ function Server:exec(fn, args, options)
             :format(utils.get_fn_location(fn)))
     end
 
+    local utils_dumps = {
+        unpack_sparse_array = string.dump(utils.unpack_sparse_array),
+    }
+
     -- The function `fn` can return multiple values and we cannot use the
     -- classical approach to work with the `pcall`:
     --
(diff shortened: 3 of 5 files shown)


luatest/utils.lua
```diff
@@ -191,4 +191,26 @@ function utils.is_tarantool_binary(path)
     return path:find('^.*/tarantool[^/]*$') ~= nil
 end
 
+function utils.unpack_sparse_array(array, index)
+    -- Embed everything so no upvalues required on server.
+    index = index or 1
+
+    local len = 0
+    for k, _ in pairs(array) do
```
ochaplashkin, line 199: Maybe it makes sense to improve the `utils.table_len()` function?

```lua
function utils.table_len(t)
    local counter = 0
    for _, _ in pairs(t) do
        counter = counter + 1
    end
    return counter
end
```

ochaplashkin (commented):
Let's change the the commit message
from
`exec: ...`
to
`server: restore sparse output support from exec`

luatest/utils.lua
DifferentialOrange: I'm not sure.

The concept of "array" is rather unclear in Lua, even though there are some tools like `ipairs` and `unpack`. The concept of "sparse array" is even more unclear. Now `table_len` is expected to return "the number of actual values in the table". 
For `{nil, 2, 3, nil, 5}` it's 3, but it's sparse array length is 5. And there is no any way for us to tell whether the developer is treating some table value as a "sparse array" and expect its length to be 5 or as a "key-value map with number keys" and expects its length to be 3. (The `{[index_id] = index}` map for a space schema is an example of key-value map with number keys. I have one in `tarantool/crud`. It always has `[0]` key, it may look like `{[0] = index_0, [2] = index_2}` if some index was removed. I'm more likely to expect that `len(index_map)` would return a count of indexes, which is the result of current `table_len`.) Since there is no way for us to choose the approach, it must be a responsibility of the developer. We may introduce `table_len` with flag to choose the approach, but I think it's more confusing than having two different functions.

(Now "sparse array len" is not a separate function since I dump the util and it's really a bother to support upvalues. But I can restore it to be a separate function in current implementation, even though in may confuse future `:exec` developers.)

DifferentialOrange:
> from exec

That's fair, updated

DifferentialOrange:
> Let's change the the commit message from `exec: ...` to `server: restore sparse output support from exec`

That's fair, updated

ochaplashkin (approved):
LGTM :rocket:

PR #355, opened 2024-02-21
Change conversion of `box.error` to string in error message assertions

CuriousGeorgiy:
This patch changes the conversion of `box.error` to string to retain the behavior after tarantool/tarantool#9105.

Needed for tarantool/tarantool#9105


luatest/server.lua
```diff
@@ -611,7 +611,9 @@ end
 
 local function exec_tail(ok, ...)
     if not ok then
-        local _ok, res = pcall(json.decode, tostring(...))
+        local err = ...
```
ylobankov, line 614: Is new code block compatible with Tarantool 1.10?

PR #361, opened 2024-03-21
Improve `luatest.log` function if a `nil` value is passed

### LINKED ISSUE #360: `cdata<void *>: NULL` can't be processed in pairs()
https://github.com/tarantool/luatest/blob/e3ea6fc459b5f9f1c4c458e05c3167f61c72d1af/luatest/log.lua#L10-L19

This function may be used in `asserts` module too:

```lua
--- Check that two values are equal.
-- <...>
function M.assert_equals(actual, expected, message, deep_analysis)
    log.info('Assert equals %s and %s', actual, expected) 
    if not comparator.equals(actual, expected) then
        failure(M.private.error_msg_equality(actual, expected, deep_analysis), message, 2)
    end
end
```
For example:

```lua
g.test_cdata = function()
    ...
    local r = server:exec(function() return nil end)
    t.assert(r, nil)
end
```
First argument `assert` function is `cdata<void *>: NULL` and the second is `nil`. This is going to cause an error: `bad argument #3 to '?' (no value)`.

It is necessary to use `select(index, ...)` for a more correct iteration of `vararg` (thx @Totktonada )

ochaplashkin:
Use `select()` to avoid `bad argument #3 to '?' (no value)` error in following example:
```lua
    luatest.log('%s and %s', cdata<void *>: NULL, nil)
```
Closes #360

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add logging to unified file (gh-324).
 - Add memory leak detection during server process execution (gh-349).
+- Improve `luatest.log` function if a `nil` value is passed (gh-360).
 
 ## 1.0.1
 
--- a/luatest/log.lua
+++ b/luatest/log.lua
@@ -11,9 +11,10 @@ local function _log(level, msg, ...)
     if not utils.version_current_ge_than(2, 5, 1) then
         return
     end
-    local args = {...}
-    for k, v in pairs(args) do
-        args[k] = pp.tostringlog(v)
+    local args = {}
+    for i = 1, select('#', ...) do
+        local v = select(i, ...)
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Add logging to unified file (gh-324).
 - Add memory leak detection during server process execution (gh-349).
+- Improve `t.log()` function if a `nil` value is passed (gh-360).
```
ylobankov, line 7: `t` -> `luatest`

ochaplashkin: Fixed :heart:

ylobankov (approved):

PR #369, opened 2024-04-11
Migrate `justrun` module and adapt it

### LINKED ISSUE #365: Add `justrun.lua` helper from tarantool to luatest and adapt it
Add the `justrun.lua` module from the tarantool repo ([tarantool/test/justrun.lua](https://github.com/tarantool/tarantool/blob/58f5ffe43bffa0901d8d577bbb4d523c4e7e682a/test/justrun.lua)) to luatest and adapt it.

ochaplashkin:
The original `justrun` module (path: tarantool/test/justrun.lua) has been moved to the current project with minor changes and will be available as follows:
```lua
    local t = require('luatest')
    t.justrun.tarantool(...)
```

This module works with the `popen` module which requires Tarantool 2.4.1 and newer. Otherwise `justrun.tarantool(dir, env, args[, opts])` will cause an error.

Closes #365

### DIFF
--- a/luatest/init.lua
+++ b/luatest/init.lua
@@ -38,6 +38,11 @@ local parametrizer = require('luatest.parametrizer')
 --
 luatest.log = require('luatest.log')
 
+--- Simple Tarantool runner and output catcher.
+--
+-- @see luatest.justrun
+luatest.justrun = require('luatest.justrun')
+
 --- Add before suite hook.
 --
 -- @function before_suite
--- a/luatest/justrun.lua
+++ b/luatest/justrun.lua
@@ -1,0 +1,172 @@
+--- Simple Tarantool runner and output catcher.
+--
+-- Sometimes it is necessary to run tarantool with particular arguments and
+-- verify its output. `luatest.server` provides a supervisor like
+-- interface: an instance is started, calls box.cfg() and we can
+-- communicate with it using net.box. Another helper in tarantool/tarantool,
+-- `test.interactive_tarantool`, aims to solve all the problems around
+-- readline console and also provides ability to communicate with the
+-- instance interactively.
+--
+-- However, there is nothing like 'just run tarantool with given args and
+-- give me its output'.
+--
+-- @module luatest.justrun
+
+local checks = require('checks')
+local fun = require('fun')
+local json = require('json')
+local fiber = require('fiber')
+
+local log = require('luatest.log')
+
+local justrun = {}
+
+local function collect_stderr(ph)
+    local f = fiber.new(function()
+        local fiber_name = "child's stderr collector"
+        fiber.name(fiber_name, {truncate = true})
+
+        local chunks = {}
+
+        while true do
+            local chunk, err = ph:read({stderr = true})
+            if chunk == nil then
+                log.warn('%s: got error, exiting: %s', fiber_name, err)
+                break
+            end
+            if chunk == '' then
+                log.info('%s: got EOF, exiting', fiber_name)
+                break
+            end
+            table.insert(chunks, chunk)
+        end
+
+        -- Glue all chunks, strip trailing newline.
+        return table.concat(chunks):rstrip()
+    end)
+    f:set_joinable(true)
+    return f
+end
+
+local function cancel_stderr_fiber(stderr_fiber)
+    if stderr_fiber == nil then
+        return
+    end
+    stderr_fiber:cancel()
+end
+
+local function join_stderr_fiber(stderr_fiber)
+    if stderr_fiber == nil then
+        return
+    end
+    return select(2, assert(stderr_fiber:join()))
+end
+
+--- Run tarantool in given directory with given environment and
+-- command line arguments and catch its output.
+--
+-- Expects JSON lines as the output and parses it into an array
+-- (it can be disabled using `nojson` option).
+--
+-- Options:
+--
+-- - nojson (boolean, default: false)
+--
+--   Don't attempt to decode stdout as a stream of JSON lines,
+--   return as is.
+--
+-- - stderr (boolean, default: false)
+--
+--   Collect stderr and place it into the `stderr` field of the
+--   return value
+--
+-- - quote_args (boolean, default: false)
+--
+--   Quote CLI arguments before concatenating them into a shell
+--   command.
+--
+-- @string dir Directory where the process will run.
+-- @tparam table env Environment variables for the process.
+-- @tparam table args Options that will be passed when the process starts.
+-- @tparam[opt] table opts Custom options: nojson, stderr and quote_args.
+-- @treturn table
+function justrun.tarantool(dir, env, args, opts)
+    checks('string', 'table', 'table', '?table')
+    opts = opts or {}
+
+    local popen = require('popen')
+
+    -- Prevent system/user inputrc configuration file from
+    -- influencing testing code.
+    env['INPUTRC'] = '/dev/null'
+
+    local tarantool_exe = arg[-1]
+    -- Use popen.shell() instead of popen.new() due to lack of
+    -- cwd option in popen (gh-5633).
+    local env_str = table.concat(fun.iter(env):map(function(k, v)
+        return ('%s=%q'):format(k, v)
+    end):totable(), ' ')
+    local args_str = table.concat(fun.iter(args):map(function(v)
+        return opts.quote_args and ('%q'):format(v) or v
+    end):totable(), ' ')
+    local command = ('cd %s && %s %s %s'):format(dir, env_str, tarantool_exe,
+                                                 args_str)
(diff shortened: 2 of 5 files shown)


luatest/justrun.lua
```diff
@@ -0,0 +1,164 @@
+local fun = require('fun')
+local json = require('json')
+local fiber = require('fiber')
+
+local log = require('luatest.log')
+
+local has_popen, popen = pcall(require, 'popen')
+if not has_popen then
+    log.error('popen module requires Tarantool 2.4.1 and newer')
```
ylobankov, line 9: Shouldn't we explicitly break here (not just to log) if Tarantool is lower than 2.4.1?

luatest/justrun.lua
```diff
@@ -0,0 +1,164 @@
...
+--   Quote CLI arguments before concatenating them into a shell
+--   command.
+--
+-- @string dir Directory where the process will run.
+-- @table env Environment variables for the process.
+-- @table args Options that will be passed when the process starts.
+-- @table opts Custom options: nojson, stderr and qoute_args.
+-- @return table
+function justrun.tarantool(dir, env, args, opts)
+    assert(type(dir) == 'string')
```
ylobankov, line 84: Is it possible to use `checks` here?

luatest/justrun.lua
```diff
@@ -0,0 +1,164 @@
...
+    return {
+        exit_code = exit_code,
+        stdout = decoded,
+        stderr = stderr,
+    }
+end
+
+-- `popen` module requires Tarantool 2.4.1 and newer
+-- Otherwise justrun module will be an empty table.
+return has_popen and justrun or {}
```
ylobankov, line 164: This should be changed if you consider my comments above.

luatest/init.lua
```diff
@@ -38,6 +38,9 @@ local parametrizer = require('luatest.parametrizer')
 --
 luatest.log = require('luatest.log')
 
+-- Simple Tarantool runner and output catcher
```
ylobankov, line 41: To be consistent with neighbor code, please change:

`--` -> `---`
Add a dot to the end of the sentence.

luatest/justrun.lua
```diff
@@ -0,0 +1,164 @@
...
+-- - quote_args (boolean, default: false)
+--
+--   Quote CLI arguments before concatenating them into a shell
+--   command.
+--
+-- @string dir Directory where the process will run.
+-- @table env Environment variables for the process.
+-- @table args Options that will be passed when the process starts.
+-- @table opts Custom options: nojson, stderr and qoute_args.
+-- @return table
```
Totktonada, line 82: What is purpose of these luadoc/ldoc comments?

I've generated the HTML pages using `ldoc -d html .` and the module is not present there at all.

If it is just for consistency with other code, then, hm. Don't know. I think that it is weird to write a comment that looks like it adheres ldoc syntax, but it isn't. Moreover, it looks like we generate and deploy the docs to GitHub Pages or some other place. But we don't.

----

I was worrying about `@table` vs `@tparam table`. I'm not sure that the former works as expected, because I had [some workarounds](https://github.com/tarantool/expirationd/blob/419cb99ca60a5c4e6fde570cd39de1c934fdc694/doc/ldoc/config.ld#L11-L13) around in the past.

BTW, it seems you're using `return` with a type, but `treturn` accepts a type.

I don't like `tparam`, `treturn` and so on keywords and prefer the short ones like `string`, `boolean` and so on. But if we're going to use them, we have to add support for them into ldoc's config/template.

luatest/init.lua
Totktonada: I'm against mimicry of ldoc comment style without actual usage of ldoc.

luatest/justrun.lua
Totktonada: As a concise alternative I propose to move the require (without pcall) to `justrun.tarantool`.

If it is called on a supported version, it works. If it is called on an unsupported version, it raises an error that `popen` is not found.

luatest/justrun.lua
```diff
@@ -0,0 +1,164 @@
+local fun = require('fun')
+local json = require('json')
+local fiber = require('fiber')
+
+local log = require('luatest.log')
```
Totktonada, line 5: I performed a little test:

```sh
./src/tarantool $(which luatest) test/app-luatest/override_panic_test.lua -v -c -l -
```

With `log`:

```lua
luatest | 2024-04-23 15:38:26.357 [19646] main/104/luatest/test.justrun I> Running a command: cd /tmp/XBCo7A && INPUTRC="/dev/null" /home/alex/projects/tarantool-meta/r/t-3/src/tarantool main.lua
```

With `luatest.log`:

```lua
luatest | 2024-04-23 15:37:39.270 [18624] main/104/luatest/luatest.log I> Running a command: "cd /tmp/1PUXlF && INPUTRC=\"/dev/null\" /home/alex/projects/tarantool-meta/r/t-3/src/tarantool main.lua"
```

As for me, it is nice to have `justrun` in the log message.

Totktonada (approved):
I have no strict objections, but, please, consider my thoughts above.

luatest/init.lua
ylobankov: Actually, we use it and publish docs to tarantool.io. Moreover, we have a [workflow](https://github.com/tarantool/luatest/actions/workflows/ldoc.yml) to test ldoc build.

Totktonada: Wow, nice. I overlooked it.

So, my nits regarding ldoc's syntax below are applicable. We should verify how it is rendered.

luatest/justrun.lua
ochaplashkin: I've moved it inside the function :rocket:

luatest/justrun.lua
ochaplashkin: Yep, I added checks via `checks`.

luatest/justrun.lua
ochaplashkin: Yep, fixed. (comment also deleted)

luatest/justrun.lua
```diff
@@ -0,0 +1,164 @@
...
+--
+-- - quote_args (boolean, default: false)
+--
+--   Quote CLI arguments before concatenating them into a shell
+--   command.
+--
+-- @string dir Directory where the process will run.
+-- @table env Environment variables for the process.
+-- @table args Options that will be passed when the process starts.
+-- @table opts Custom options: nojson, stderr and qoute_args.
```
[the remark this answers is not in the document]
ochaplashkin: Fixed.

luatest/justrun.lua
ochaplashkin: I agree with you and it's my bad (didn't render the documentation). Now it looks as follows:

изображение

luatest/init.lua
ochaplashkin: Fixed. See my comment below to see how it looks in the documentation.

luatest/justrun.lua
ochaplashkin: Yes, this became obvious to me only after some time. Because we override the `log` function in the `luatest.log` module, we will always get messages like:

```lua
luatest | ... [...] main/104/luatest/luatest.log I> ...
```

Obviously, the `luatest.log` module needs improvements.

luatest/justrun.lua
```diff
@@ -0,0 +1,172 @@
+--- Simple Tarantool runner and output catcher.
+--
+-- Sometimes it is necessary to run tarantool with particular arguments and
+-- verify its output. `luatest.server` provides a supervisor like
+-- interface: an instance is started, calls box.cfg() and we can
+-- communicate with it using net.box. Another helper,
+-- `test.interactive_tarantool`, aims to solve all the problems around
```
ylobankov, line 7: I think we should give a clue what is this module. So I suggest this:

```
-- communicate with it using net.box. Another helper in tarantool/tarantool,
-- `test.interactive_tarantool`, aims ...
```

ylobankov (approved):

luatest/justrun.lua
ochaplashkin: Fixed.

```
diff --git a/luatest/justrun.lua b/luatest/justrun.lua
index 391fabf..d21ef8d 100644
--- a/luatest/justrun.lua
+++ b/luatest/justrun.lua
@@ -3,7 +3,7 @@
 -- Sometimes it is necessary to run tarantool with particular arguments and
 -- verify its output. `luatest.server` provides a supervisor like
 -- interface: an instance is started, calls box.cfg() and we can
--- communicate with it using net.box. Another helper,
+-- communicate with it using net.box. Another helper in tarantool/tarantool,
 -- `test.interactive_tarantool`, aims to solve all the problems around
 -- readline console and also provides ability to communicate with the
 -- instance interactively.
 ```

PR #377, opened 2024-05-23
Make assert_error_covers support stacked diagnostics

### LINKED ISSUE #372: `assert_error_covers` doesn't unpack `prev`
An error object (created with `box.error`) may contain a cause (`prev`), which points to another error object. It isn't unpacked by `assert_error_covers` (added in #362), which makes the helper useless for checking stacked error diagnostics.

nshy:
That is if there is `expected.prev` then it should cover `actual.prev` and so on recursively. Also make existing tests for `assert_error_covers` more precise.

Closes #372

### DIFF
--- a/luatest/assertions.lua
+++ b/luatest/assertions.lua
@@ -648,10 +648,39 @@ function M.assert_error_msg_matches(pattern, fn, ...)
     end
 end
 
+-- If it is box.error that unpack it recursively. If it is not then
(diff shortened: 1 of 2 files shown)


locker (approved):

ochaplashkin (approved):
Good tests, thanks! :rocket:

ylobankov (approved):

PR #378, opened 2024-05-27
Add --list-test-cases and --run-test-case

Totktonada:
These options allow an external runner (such as [test-run][test-run]) to run the test cases separately: isolated and/or parallel.

The idea is the following.

* test-run has some criteria, whether to parallelize the given test.
* If it needs to be parallelized, test-run invokes `luatest --list-test-cases <..path to test..>`.
* Then each test case is run in its own process using `luatest --run-test-case <..test case..> <..path to test..>`.

The `--run-test-case` option may be simulated using `--pattern`, but it requires regexp escaping. It is more convenient to have a separate option to pass a verbatim test case name.

The new CLI options are mostly for scripting and not for interactive use. So, no short option are added for them.

[test-run]: https://github.com/tarantool/test-run

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Changed error message for too long Unix domain socket paths (gh-341).
 - Add `cbuilder` helper as a declarative configuration builder (gh-366).
 - Make `assert_error_*` additionally check error trace if required.
+- Add `--list-test-cases` and `--run-test-case` CLI options.
 
 ## 1.0.1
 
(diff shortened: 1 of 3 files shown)


Totktonada:
See https://github.com/tarantool/test-run/pull/433 for a usage example.

ochaplashkin (approved):

Totktonada:
Bump of the submodule in test-run: https://github.com/tarantool/test-run/pull/436.

PR #383, opened 2024-07-09
Add `assert_items_exclude` matcher

ImeevMA:
This patch adds the `assert_items_exclude(actual, expected, message)` matcher. This matcher can be useful, for example, to verify that gc is working correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Add alias `--no-capture` for the option `-c` (gh-391).
 - Fix reporting of an assertion failure in `Server:exec()` in case verbose
   error serialization is enabled in Tarantool (gh-376).
+- Added `assert_items_exclude`.
 
 ## 1.0.1
 
--- a/luatest/assertions.lua
+++ b/luatest/assertions.lua
@@ -411,6 +411,21 @@ function M.assert_items_include(actual, expected, message)
     end
 end
 
+--- Checks that one table does not include any items of another, irrespective of their keys.
+--
+-- @param actual
+-- @param expected
+-- @string[opt] message
+function M.assert_items_exclude(actual, expected, message)
+    if type(actual) ~= 'table' or type(expected) ~= 'table' then
+        failure('Argument 1 and 2 must be tables', nil, 2)
+    end
+    if not comparator.are_disjoint(expected, actual) then
+        expected, actual = prettystr_pairs(expected, actual)
+        fail_fmt(2, message, 'Expected no item values from: %s\nTo be present in: %s', expected, actual)
+    end
+end
+
 local function table_slice(actual, expected)
     if type(expected) ~= 'table' or type(actual) ~= 'table' then
         return actual
--- a/luatest/comparator.lua
+++ b/luatest/comparator.lua
@@ -70,6 +70,23 @@ function comparator.is_subset(actual, expected)
     return #expected_array - found_count
 end
 
+-- Returns false if 'actual' or 'expected' are not tables or their value sets
+-- intersect. Returns true otherwise.
+function comparator.are_disjoint(actual, expected)
+    if (type(actual) ~= 'table') or (type(expected) ~= 'table') then
+        return false
+    end
+
+    for _, a in pairs(actual) do
+        for _, b in pairs(expected) do
+            if comparator.equals(a, b) then
+                return false
+            end
+        end
+    end
+    return true
+end
+
 -- This is a specialized metatable to help with the bookkeeping of recursions
 -- in table_equals(). It provides an __index table that implements utility
 -- functions for easier management of the table. The "cached" method queries
--- a/test/luaunit/assertions_test.lua
+++ b/test/luaunit/assertions_test.lua
@@ -333,6 +333,39 @@ function g.test_assert_items_include()
     assert_failure(subject, {1,2,3}, {1,1,2,3})
 end
 
+function g.test_assert_items_exclude()
+    local subject = t.assert_items_exclude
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Changelog
+- Added `assert_items_exclude`.
```
locker, line 2: The changelog entry should be appended to the end of the "Unrelased" section.

luatest/assertions.lua
```diff
@@ -411,6 +411,21 @@ function M.assert_items_include(actual, expected, message)
...
 end
 
+--- Checks that one table does not include any items of another, irrespective of their keys.
+--
+-- @param actual
+-- @param expected
+-- @string[opt] message
+function M.assert_items_exclude(actual, expected, message)
+    if type(actual) ~= 'table' or type(expected) ~= 'table' then
+        failure('Argument 1 and 2 must be tables', nil, 3)
```
locker, line 421: Should be 2, not 3 (it's the stack level).

Anyway, we don't have such a check in `assert_items_include`. Do we really want it here?

luatest/comparator.lua
```diff
@@ -70,6 +70,31 @@ function comparator.is_subset(actual, expected)
     return #expected_array - found_count
 end
 
+-- Checks that any element from the actual is an element from the expected.
+function comparator.any(actual, expected)
```
locker, line 74: The function should probably be called `intersects`.

luatest/comparator.lua
```diff
@@ -70,6 +70,31 @@ function comparator.is_subset(actual, expected)
...
+    if (type(actual) ~= 'table') or (type(expected) ~= 'table') then
+        return false
+    end
+
+    local expected_array = {}
+    local expected_casted = {}
+    for _, v in pairs(expected) do
+        table.insert(expected_array, v)
+    end
```
locker, line 83: Why do you convert `expected` to an array? In `is_subset` this is used to return the number of elements that are present in `expected` but not in `actual` but here don't account it.

`comparator.cast` used below doesn't seem to be necessary, either.

test/luaunit/assertions_test.lua
```diff
@@ -333,6 +333,27 @@ function g.test_assert_items_include()
...
+
+    assert_failure(subject, nil, {1})
+    assert_failure(subject, {}, nil)
+    assert_failure(subject, {1,2,3}, {1,2,3,4})
+    assert_failure(subject, {1,2,3}, {1,1,2,3})
+
+    subject({}, {1})
+    subject({},{})
+    subject({nil},{nil})
+    subject({1},{})
```
locker, line 354: Please add more success cases: maps with keys, maps with more than one item, ...

locker (changes requested):

CHANGELOG.md
ImeevMA: Thanks, fixed.

luatest/assertions.lua
ImeevMA: True, removed. I must be thought that the message looks strange in case we have non-table argument, but since message in `assert_items_include` is quite similar, it is better to remove this check.

luatest/comparator.lua
ImeevMA: Reworked the function a bit. It is now called `are_disjoint()` and returns `false` if at least one of the arguments is not a table or if there is a common element for both arguments. Otherwise it returns `true`.

luatest/comparator.lua
ImeevMA: Thanks, fixed.

test/luaunit/assertions_test.lua
ImeevMA: Added.

luatest/comparator.lua
```diff
@@ -70,6 +70,22 @@ function comparator.is_subset(actual, expected)
     return #expected_array - found_count
 end
 
+-- Checks that any element from the actual is an element from the expected.
+function comparator.are_disjoint(actual, expected)
+    if (type(actual) ~= 'table') or (type(expected) ~= 'table') then
+        return false
```
locker, line 76: A scalar value is disjoint (doesn't intersect) with any other table or scalar value, no? If you want to explicitly forbid using this function with scalar values, better use an explicit check IMO, like the one you had in the previous version of this patch, otherwise this helper function, as well as the error message, will look confusing:

```
Expected no item values from: 10
To be present in: [1, 2, 3]
```

test/luaunit/assertions_test.lua
```diff
@@ -333,6 +333,32 @@ function g.test_assert_items_include()
...
+    subject({one=1},{})
+    subject({},{one=1})
+    subject({1, 2, 3},{4, 5, 6, 7})
+    subject({one=1, two=2, three=3},{four=4, five=5})
+    subject({one=1, 2, 3},{four=4, 5})
```
locker, line 359: Please add more checks with scalar values, like `subject(1, {1})`, `subject(1, 1)`, etc.

luatest/comparator.lua
ImeevMA: Fixed. Added explicit check.

test/luaunit/assertions_test.lua
ImeevMA: Added a few tests with scalar values.

locker (approved):

Totktonada (approved):
No objections from my side. Thanks!

Totktonada:
`assert_items_disjoint` maybe is a more natural name.

I'm OK with the current `assert_items_exclude`, though.

PR #392, opened 2024-09-27
Add a long option --no-capture

### LINKED ISSUE #391: Add a long-option for no-capture mode
For unknown reasons, there is no long option for no-capture mode, only short-option (`-c`).
I propose to fix this.

ligurio:
The option `-c` allows to disable capture and this option has no long variant with self-explained name.
The patch fixes that.

Closes #391


Totktonada (approved):
I like the idea!

ylobankov (approved):

PR #398, opened 2024-10-22
Propagate stack trace from Server:exec()

### LINKED ISSUE #396: No stack trace on error/fail in `server:exec()`
Luatest prints the stack trace if the test fails an assertion or raises an error. This is really useful in case a failure happens in a function, which is called in more than one place in the test, because in this case the assertion/error itself will point to the code inside the function so that without the stack trace one can't figure out where the test failed.

To illustrate this issue, here's a test.

<details><summary>test file</summary>

```lua
local t = require('luatest')
local server = require('luatest.server')

local g = t.group()

g.before_all(function(cg)
    cg.server = server:new()
    cg.server:start()
end)

g.after_all(function(cg)
    cg.server:drop()
end)

g.test_fail_local = function()
    local function check(x)
        t.assert_ge(x, 1)
        t.assert_le(x, 5)
    end
    check(1)
    check(10)
end

g.test_fail_server = function(cg)
    cg.server:exec(function()
        local function check(x)
            t.assert_ge(x, 1)
            t.assert_le(x, 5)
        end
        check(1)
        check(10)
    end)
end

g.test_error_local = function()
    local function check(x)
        if x < 1 or x > 5 then
            error('invalid x', 0)
        end
    end
    check(1)
    check(10)
end

g.test_error_server = function(cg)
    cg.server:exec(function()
        local function check(x)
            if x < 1 or x > 5 then
                error('invalid x', 0)
            end
        end
        check(1)
        check(10)
    end)
end
```

</details>

Put this test somewhere in the Tarantool repository and run. You'll get the following output.

<details><summary>luatest output</summary>

```
Tests with errors:
------------------

1) box-luatest.test.test_error_local
invalid x
stack traceback:
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:38: in function 'check'
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:42: in function 'box-luatest.test.test_error_local'
        ...
        [C]: in function 'xpcall'
2) box-luatest.test.test_error_server
invalid x
stack traceback:
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:46: in function 'box-luatest.test.test_error_server'
        ...
        [C]: in function 'xpcall'
artifacts:
        server -> /tmp/t/artifacts/server-cU_eIc40j6IC

Failed tests:
-------------

1) box-luatest.test.test_fail_local
...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:18: Assertion failed: 10 <= 5
stack traceback:
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:18: in function 'check'
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:21: in function 'box-luatest.test.test_fail_local'
        ...
        [C]: in function 'xpcall'
2) box-luatest.test.test_fail_server
...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:28: Assertion failed: 10 <= 5
stack traceback:
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:25: in function 'box-luatest.test.test_fail_server'
        ...
        [C]: in function 'xpcall'
artifacts:
        server -> /tmp/t/artifacts/server-cU_eIc40j6IC
```

</details>

Note that when the `check()` function is called locally the stack trace can be used to pinpoint the call location:

```
stack traceback:
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:18: in function 'check'
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:21: in function 'box-luatest.test.test_fail_local'

stack traceback:
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:18: in function 'check'
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:21: in function 'box-luatest.test.test_fail_local'
```

If `check()` is called from `server:exec()`, the stack trace points to the line where `server:exec()` is located:

```
stack traceback:
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:46: in function 'box-luatest.test.test_error_server'

stack traceback:
        ...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:25: in function 'box-luatest.test.test_fail_server'
```

If the test fails an assertion, we can at least see the line where the assertion is located because it seems to be attached to the assertion error object. This isn't enough in our case but it's something:

```
...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:18: Assertion failed: 10 <= 5

...d/src/tarantool/tarantool/test/box-luatest/test_test.lua:28: Assertion failed: 10 <= 5
```

If the test fails with an error, we lack even this little information (fat least for raw Lua errors; Box errors are enriched with `file:line` but they don't always point where they should).

locker:
If a function executed with `Server:exec()` fails, luatest prints the stack trace only up to the `Server:exec()` call. If the failure happens to be in a nested function, this makes it impossible to figure out what happened.

Let's propagate the error stack trace from the server to the runner and make the runner concatenate it with its own stack trace. To achieve that, we wrap errors raised by the function executed by `Serer:exec()` to save the trace.

Closes #396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Fix reporting of an assertion failure in `Server:exec()` in case verbose
   error serialization is enabled in Tarantool (gh-376).
 - Added `assert_items_exclude`.
+- Strip useless `...` lines from error trace.
 
 ## 1.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
   error serialization is enabled in Tarantool (gh-376).
 - Added `assert_items_exclude`.
 - Strip useless `...` lines from error trace.
+- Fix error trace reporting for functions executed with `Server:exec()`
+  (gh-396).
 
 ## 1.0.1
 
--- a/luatest/utils.lua
+++ b/luatest/utils.lua
@@ -61,22 +61,46 @@ end
 
 -- Check if line of stack trace comes from inside luatest.
 local function is_luatest_internal_line(s)
-    return s:find('[/\\]luatest[/\\]') or s:find('bin[/\\]luatest')
+    if s:find('bin[/\\]luatest') then
+        return true
+    end
+    if s:find('[/\\]luatest[/\\]') then
+        -- Do not strip lines originating from luatest test files because
+        -- we want to see stack traces when we test luatest.
+        return not s:find('[/\\]luatest[/\\]test[/\\]')
+    end
+    return false
 end
 
 function utils.strip_luatest_trace(trace)
     local lines = trace:split('\n')
-    local result = {lines[1]} -- always keep 1st line
+
+    -- Scan the stack trace backwards (from caller to callee) and strip all
+    -- frames related to luatest, as well as `[C]:`, `...`, `eval:` frames
+    -- called by them because they don't change context.
+    local result = {}
     local keep = true
-    for i = 2, table.maxn(lines) do
+    for i = table.maxn(lines), 2, -1 do
         local line = lines[i]
-        -- `[C]:` lines don't change context
-        if not line:find('^%s+%[C%]:') then
+        if not (line:find('^%s+%[C%]:') or
+                line:find('^%s+%.%.%.$') or
+                line:find('^%s+eval:')) then
             keep = not is_luatest_internal_line(line)
         end
         if keep then
             table.insert(result, line)
         end
+    end
+
+    -- Always keep the 1st line because it's the header ('stack traceback:').
+    table.insert(result, lines[1])
+
+    -- Since we scanned the stack trace backwards, we need to reverse
+    -- the result.
+    for i = 1, math.floor(#result / 2) do
+        local v = result[#result - i + 1]
+        result[#result - i + 1] = result[i]
+        result[i] = v
     end
     return table.concat(result, '\n')
 end
--- a/luatest/assertions.lua
+++ b/luatest/assertions.lua
@@ -164,12 +164,26 @@ local function pcall_check_trace(level, fn, ...)
     return ok, err
 end
 
+--
+-- Wrapper around pcall that:
+-- * checks the stack trace for box errors;
+-- * unwraps the error if it was wrapped by Server:exec().
+--
+local function pcall_wrapper(level, fn, ...)
+    local ok, err = pcall_check_trace(level + 1, fn, ...)
+    if not ok and type(err) == 'table' and
+            err.class == 'LuatestErrorWrapper' then
+        err = err.error
+    end
+    return ok, err
+end
+
 --- Check that calling fn raises an error.
 --
 -- @func fn
 -- @param ... arguments for function
 function M.assert_error(fn, ...)
-    local ok, err = pcall_check_trace(2, fn, ...)
+    local ok, err = pcall_wrapper(2, fn, ...)
     if ok then
         failure("Expected an error when calling function but no error generated", nil, 2)
     end
@@ -565,7 +579,7 @@ function M.assert_str_matches(value, pattern, start, final, message)
 end
 
 local function _assert_error_msg_equals(stripFileAndLine, expectedMsg, func, ...)
-    local no_error, error_msg = pcall_check_trace(3, func, ...)
+    local no_error, error_msg = pcall_wrapper(3, func, ...)
     if no_error then
         local failure_message = string.format(
             'Function successfully returned: %s\nExpected error: %s',
@@ -631,7 +645,7 @@ end
 -- @func fn
 -- @param ... arguments for function
 function M.assert_error_msg_contains(expected_partial, fn, ...)
-    local no_error, error_msg = pcall_check_trace(2, fn, ...)
+    local no_error, error_msg = pcall_wrapper(2, fn, ...)
     log.info('Assert error message %s contains %s', error_msg, expected_partial)
     if no_error then
         local failure_message = string.format(
@@ -654,7 +668,7 @@ end
 -- @func fn
 -- @param ... arguments for function
 function M.assert_error_msg_matches(pattern, fn, ...)
-    local no_error, error_msg = pcall_check_trace(2, fn, ...)
+    local no_error, error_msg = pcall_wrapper(2, fn, ...)
     if no_error then
         local failure_message = string.format(
             'Function successfully returned: %s\nExpected error matching: %s',
@@ -694,7 +708,7 @@ end
 -- @func fn
 -- @param ... arguments for function
 function M.assert_error_covers(expected, fn, ...)
-    local ok, actual = pcall_check_trace(2, fn, ...)
+    local ok, actual = pcall_wrapper(2, fn, ...)
     if ok then
         fail_fmt(2, nil,
                  'Function successfully returned: %s\nExpected error: %s',
--- a/luatest/runner.lua
+++ b/luatest/runner.lua
@@ -474,6 +474,16 @@ function Runner.mt:protected_call(instance, method, pretty_name)
     end, function(e)
         -- transform error into a table, adding the traceback information
         local trace = debug.traceback('', 3):sub(2)
+        if type(e) == 'table' and e.class == 'LuatestErrorWrapper' then
+            -- This is an error wrapped by Server:exec() to save the trace.
+            -- Concatenate the current trace with the saved trace and restore
+            -- the original error.
+            assert(e.trace ~= nil)
+            assert(e.error ~= nil)
+            trace = e.trace .. '\n' ..
+                    trace:sub(string.len('stack traceback:\n') + 1)
+            e = e.error
+        end
         if utils.is_luatest_error(e) then
             return {status = e.status, message = e.message, trace = trace}
         else
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -823,6 +823,10 @@ function Server:exec(fn, args, options)
         error(('bad argument #3 for exec at %s: an array is required'):format(utils.get_fn_location(fn)))
     end
 
+    -- Note that we wrap any error raised by the executed function to save
+    -- the original trace. It will be used by the runner to report the full
+    -- error trace, see Runner:protected_call().
+    --
     -- If the function fails an assertion in an open transaction, Tarantool
     -- will raise the "Transaction is active at return from function" error,
     -- thus overwriting the original error raised by the assertion. To avoid
@@ -836,12 +840,19 @@ function Server:exec(fn, args, options)
                 debug.setupvalue(fn, i, require(passthrough_ups[name]))
             end
         end
-        local result
-        if args == nil then
-            result = {pcall(fn)}
-        else
-            result = {pcall(fn, unpack(args))}
-        end
+        local result = {xpcall(function()
+            if args == nil then
+                return fn()
+            else
+                return fn(unpack(args))
+            end
+        end, function(e)
+            return {
+                class = 'LuatestErrorWrapper',
+                error = e,
+                trace = debug.traceback('', 3):sub(2),
+            }
+        end)}
         if not result[1] then
             box.rollback()
         end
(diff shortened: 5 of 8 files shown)


luatest/utils.lua
```diff
@@ -61,23 +61,33 @@ end
...
         if keep then
             table.insert(result, line)
         end
     end
+    table.insert(result, lines[1]) -- always keep 1st line
+    for i = 1, math.floor(#result / 2) do
+        local v = result[#result - i + 1]
+        result[#result - i + 1] = result[i]
+        result[i] = v
+    end
```
Totktonada, line 90: A comment like 'Reverse the list' would be helpful.

luatest/utils.lua
```diff
@@ -61,23 +61,33 @@ end
...
-        if not line:find('^%s+%[C%]:') then
+        -- `[C]:`, `...`, `eval:` lines don't change context
+        if not (line:find('^%s+%[C%]:') or
+                line:find('^%s+%.%.%.$') or
+                line:find('^%s+eval:')) then
             keep = not is_luatest_internal_line(line)
         end
         if keep then
             table.insert(result, line)
         end
```
Totktonada, line 83: The logic is not very clear here. As I understood the idea is to strip all the frames that originate from luatest and their corresponding `[C]` and similar frames.

However, we start stripping from the bottom all the frames (`keep = false`) without checking, whether they correspond to luatest. Why?

If my understanding above is correct, we want to keep non-luatest `[C]` and other frames. Shouldn't we check whether the line correspond to luatest first and then check for `[C]`?

At least, some clarification would be helpful.

Totktonada (commented):
Questions for the first commit.

luatest/utils.lua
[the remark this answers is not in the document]
locker: Makes sense. Added some comments. Hope it makes the code clear.

<details><summary>diff</summary>

```diff
diff --git a/luatest/utils.lua b/luatest/utils.lua
index ce269ada28ec..014179e700e2 100644
--- a/luatest/utils.lua
+++ b/luatest/utils.lua
@@ -61,18 +61,27 @@ end
 
 -- Check if line of stack trace comes from inside luatest.
 local function is_luatest_internal_line(s)
-    return s:find('bin[/\\]luatest') or
-        (s:find('[/\\]luatest[/\\]') and
-         not s:find('[/\\]luatest[/\\]test[/\\]'))
+    if s:find('bin[/\\]luatest') then
+        return true
+    end
+    if s:find('[/\\]luatest[/\\]') then
+        -- Do not strip lines originating from luatest test files because
+        -- we want to see stack traces when we test luatest.
+        return not s:find('[/\\]luatest[/\\]test[/\\]')
+    end
+    return false
 end
 
 function utils.strip_luatest_trace(trace)
     local lines = trace:split('\n')
+
+    -- Scan the stack trace backwards (from caller to callee) and strip all
+    -- frames related to luatest, as well as `[C]:`, `...`, `eval:` frames
+    -- called by them because they don't change context.
     local result = {}
-    local keep = false
+    local keep = true
     for i = table.maxn(lines), 2, -1 do
         local line = lines[i]
-        -- `[C]:`, `...`, `eval:` lines don't change context
         if not (line:find('^%s+%[C%]:') or
                 line:find('^%s+%.%.%.$') or
                 line:find('^%s+eval:')) then
@@ -82,7 +91,12 @@ function utils.strip_luatest_trace(trace)
             table.insert(result, line)
         end
     end
-    table.insert(result, lines[1]) -- always keep 1st line
+
+    -- Always keep the 1st line because it's the header ('stack traceback:').
+    table.insert(result, lines[1])
+
+    -- Since we scanned the stack trace backwards, we need to reverse
+    -- the result.
     for i = 1, math.floor(#result / 2) do
         local v = result[#result - i + 1]
         result[#result - i + 1] = result[i]
```

</summary>

luatest/utils.lua
```diff
@@ -61,23 +61,33 @@ end
...
     end
+    table.insert(result, lines[1]) -- always keep 1st line
```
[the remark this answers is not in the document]
locker: Added. See the diff [here](https://github.com/tarantool/luatest/pull/398#discussion_r1816484034).

luatest/utils.lua
locker: Added. See the diff [here](https://github.com/tarantool/luatest/pull/398#discussion_r1816484034).

luatest/utils.lua
locker: Actually, `keep` should be `true` initially, thanks for noticing. Fixed it. This doesn't change the output because a stack trace starts with `bin/luatest`.

test/server_test.lua
```diff
@@ -208,10 +208,11 @@ g.test_net_box_exec = function()
 
     do
         local _l_fn, fn = l(), function() error('X_x') end
-        t.assert_error_msg_equals(
-            efmt(_l_fn, 'X_x'),
-            exec, fn
-        )
```
Totktonada, line 214: It looks dangerous. Shouldn't this helper understands luatest's wrapping and unwrap it automatically?

The same regarding `t.assert_error_msg_contains` and other similar error capturing methods.

I guess that this behavior change may hit many tests.

Totktonada: > \- Make `assert_error_*` ignore errors raised by assertion failures.

I would say 're-raise' instead of 'ignore' here to make the new behavior more obvious from the text.

----

The idea to re-raise assertions is interesting.

It seems safe in context of `t.assert_error_msg_*{equals,contains,matches}`, because AFAIS a luatest error table has no `__tostring` metamethod and these functions never accept these errors before.

The things are different for `t.assert_error` (and maybe for `t.assert_error_covers`). For example, `test_bs_leader_not_goes_ro_if_appointed` from https://github.com/tarantool/tarantool-ee/pull/863 has a retrying helper, which uses `t.assert_equals` in its predicate (to get a readable errors in case of a failure), and the test case verifies that the retrying fails using `t.assert_error`.

Also, sometimes I use a plain `assert` to verify test case preconditions and use `t.assert_*` to verify what the test case aims to check. In this situation the error from the plain assert is more severe.

I'm in some doubt here.

I propose to discuss other possible solutions. What comes into my head:

* Tune `t.assert_error_msg_*{equals,contains,matches}` to look into the `message` field of the error if it is a table.
* Rename a general-looking `{class = 'LuatestError', status = 'error', <...>}` to something like `{class = 'LuatestError', status = 'wrapped_by_exec', <...>}` and unwrap it in `t.assert_error_*` before any comparisons.

Or maybe I need to spin it in the head a bit more. For a while, just wrote down my current thoughts.

locker: > The things are different for `t.assert_error` (and maybe for `t.assert_error_covers`). For example, `test_bs_leader_not_goes_ro_if_appointed` from [tarantool/tarantool-ee#863](https://github.com/tarantool/tarantool-ee/pull/863) has a retrying helper, which uses `t.assert_equals` in its predicate (to get a readable errors in case of a failure), and the test case verifies that the retrying fails using `t.assert_error`.

If you meant this

```lua
    -- In case of long yield the test may finish as a
    -- false positive. Though it's very unlikely.
    t.assert_error(wait_database_mode, g, 'instance-001', 'ro')
```

I would rewrite that test because as it is now it would pass even if the tested code throws a different kind of error (e.g. a Lua assertion or syntax error). IMO we should raise a proper error in this test and check that the expected error is raised. Actually, I'd deprecate `t.assert_error` altogether for this reason - it's too dangerous. I myself never use it and most Tarantool developers don't, either:

```
# Ran in Tarantool repo:
$ grep -r '\<assert_error\>' test | wc -l
15
```

> Also, sometimes I use a plain `assert` to verify test case preconditions and use `t.assert_*` to verify what the test case aims to check. In this situation the error from the plain assert is more severe.

Why? Luatest assertions are much more powerful and result in much more verbose error messages.

> I propose to discuss other possible solutions. What comes into my head:
> 
> * Tune `t.assert_error_msg_*{equals,contains,matches}` to look into the `message` field of the error if it is a table.
> * Rename a general-looking `{class = 'LuatestError', status = 'error', <...>}` to something like `{class = 'LuatestError', status = 'wrapped_by_exec', <...>}` and unwrap it in `t.assert_error_*` before any comparisons.

We could do that or we could just rewrite `pcall_wrapper`introduced in this patch  to only unwrap `status = 'error'` errors (they aren't normally raised by luatest), without re-raising failures. However, I prefer to make `t.assert_error_*` let assertion failures go through because:
- Catching assertion failures and checking their message in tests makes no sense.
- Catching assertion failures with `t.assert_error` is dangerous because it catches all kinds of errors.
- If we let `t.assert_error_*` catch an assertion failure (as it does now), we'd get a wrong error and trace: instead of the original assertion failure, we'd see something like "Error message expected: ... Error message received: expected: foo actual: bar", which is confusing.

Totktonada: Sounds meaningful, thanks for the clarification.

I'm OK, but, please, write the new rules of the game somewhere: in README, for example. It shouldn't be an implicit knowledge.

> Actually, I'd deprecate `t.assert_error` altogether for this reason - it's too dangerous.

It feels valid for me too. At least, I would write a recommendation to wait certain errors and discourage `t.assert_error` or a catching-all `pcall` use. Fits good to README as well, IMHO.

----

> I would rewrite that test because as it is now it would pass even if the tested code throws a different kind of error (e.g. a Lua assertion or syntax error). IMO we should raise a proper error in this test and check that the expected error is raised.

I agree. `t.assert_equals` is attractive, because I don't need to bother with constructing a good error message. OTOH, it seems good to write a message that is more specific for this retrying helper.

@georgiy-belyanin Please, take a look on our discussion.

----

Disregarding of the main discussion topic `status = 'error'` feels too general and we can name it in a somehow more specificaly to the exec's wrapping. Or clarify somewhere in the code, what does `status = 'error'` means and when it should/shouldn't be used.

I guess you have an idea about some `assert`/`test status`/`error` separation, but it seems quite implicit now. At least, I'm not familiar with luatest's code enough.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
   error serialization is enabled in Tarantool (gh-376).
 - Added `assert_items_exclude`.
 - Strip useless `...` lines from error trace.
+- Make `assert_error_*` ignore errors raised by assertion failures.
```
Totktonada, line 24: Nit: I would say 're-raise' instead of 'ignore', because the latter can mean 'suppress'.

Totktonada (approved):

test/server_test.lua
locker: > * If we let `t.assert_error_*` catch an assertion failure (as it does now), we'd get a wrong error and trace: instead of the original assertion failure, we'd see something like "Error message expected: ... Error message received: expected: foo actual: bar", which is confusing.

Come to think of it, my patch actually does overwrite the assertion failure trace by re-raising the error :(

So I think I'll go with this proposition of your for now and won't re-raise assertion failure:

> * Rename a general-looking `{class = 'LuatestError', status = 'error', <...>}` to something like `{class = 'LuatestError', status = 'wrapped_by_exec', <...>}` and unwrap it in `t.assert_error_*` before any comparisons.

CHANGELOG.md
locker: Dropped this patch. Now I wrap exec error with a special error class and unwrap only those errors in `t.assert_error_*`, as you initially suggested. Please take another look.

test/server_test.lua
locker: > Disregarding of the main discussion topic `status = 'error'` feels too general and we can name it in a somehow more specificaly to the exec's wrapping. Or clarify somewhere in the code, what does `status = 'error'` means and when it should/shouldn't be used.
> 
> I guess you have an idea about some `assert`/`test status`/`error` separation, but it seems quite implicit now. At least, I'm not familiar with luatest's code enough.

Agree, I introduced a new error class for errors wrapped by `Server:exec()`. Don't know why I decided to reuse luatest status in the first place - it does look confusing.

Totktonada (approved):

PR #399, opened 2024-10-29
Logging improvements

### LINKED ISSUE #382: reduce, refactor logs
Now if I run a cluster that contains: 2 routers, 3 storages, 1 manager, etc I see all logs.
But all tests runs only one host! I want watching only test-flow from the host.

Note: From time to time I have 60-100 screens log-text about ONE FAILED TEST.

So, It would be nice:

1. Hide all logs that generated by not luatest host
   - You can provide an option to show them, if a user wants
2. Solve logs, that required `:exec`:
   - user must know about there are errors in them WITHOUT looking through logs prefixed as `router 1 |`, `router 2 |` etc
3. All logs (that stores to /tmp/t) MUST NOT contain 'color' ESC-sequences (i.e. be grepable, printable through html, etc)

PS:

if I do 

```
$ make test
```

and one test is failed, I want a mode that showes only reduced pass-failed log without spam about all problems and happen things in cluster that contains 10-20 nodes.

Please, provide an option to see what test is ok and what is failed. Let's full log be placed somewhere but not showed in terminal!

locker:
The goal of this PR is to fix spamming stdout on test failure with unified logging is enabled. Apart from that, it also brings a few improvements to unified logging.

Close #382

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 - Add logging to unified file (gh-324).
 - Add memory leak detection during server process execution (gh-349).
-- Improve `luatest.log` function if a `nil` value is passed (gh-360).
 - Added `assert_error_covers`.
 - Add more logs (gh-326).
 - Add `justrun` helper as a tarantool runner and output catcher (gh-365).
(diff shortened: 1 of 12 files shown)


Totktonada (approved):
If I understood the changes right, after the patch we have the following situation: we have `-c` to see extra logs (child processes output, invoked assertions) and `-l` to write them to the file. The default is to drop these extra logs. It should be convenient, I think.

I'm not in context of the work that was done for unified logging before, so my review is not deep. However, the proposed patches look meaningful for me.

PR #403, opened 2024-11-13
fix: calculation of coverage percentage

a1div0:
The following small pull request fixes the issue with calculating the test coverage percentage. To achieve this, instead of initializing with an empty statistics list, it is populated with all Lua files from the LUATESTLUACOVROOT directory (by default, this is the project directory).

The CHANGELOG has been updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Fix error trace reporting for functions executed with `Server:exec()`
   (gh-396).
 - Remove pretty-printing of `luatest.log` arguments.
+- Fixed calculation of coverage percentage (gh-402)
 
 ## 1.0.1
 
--- a/luatest/coverage_utils.lua
+++ b/luatest/coverage_utils.lua
@@ -29,6 +29,17 @@ local function with_cwd(dir, fn)
     assert(fio.chdir(old), 'Failed to chdir to ' .. old)
 end
 
+local function find_luas(list_of_lua_modules, path)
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - Fix error trace reporting for functions executed with `Server:exec()`
   (gh-396).
 - Remove pretty-printing of `luatest.log` arguments.
+- Fixed calculation of coverage percentage (gh-402)
```
locker, line 26: Bad changelog - unclear what exactly was wrong. Would be nice to say that it added uncovered files to the coverage report.

Also, please write a descriptive commit message. It should include `Closes #XXX`. We don't start subject lines with "fix: ...". Should probably be "Add uncovered files to coverage report".

luatest/coverage_utils.lua
```diff
@@ -42,6 +53,10 @@ function export.enable()
         for _, item in pairs(export.DEFAULT_EXCLUDE) do
             table.insert(config.exclude, item)
         end
+
+        runner.data = {}
+        find_luas(runner.data, root)
```
locker, line 58: Accessing undocumented fields of `luacov.runner`. looks dangerous. At least, this requires a comment.

The only alternative I see is writing a coverage report file with zero counters before initializing `luacov.runner`. This could be done with `luacov.stats.save`.

@Totktonada  WDYT?

a1div0:
Sorry, this has already been fixed [here](https://github.com/lunarmodules/luacov/pull/71)

PR #406, opened 2024-12-15
Migrate cluster module and adapt it

### LINKED ISSUE #368: Add `cluster.lua` module from tarantool to luatest and adapt it
Add the `cluster.lua` module from the tarantool repo ([tarantool/test/cluster.lua](https://github.com/tarantool/tarantool/blob/58f5ffe43bffa0901d8d577bbb4d523c4e7e682a/test//config-luatest/cluster.lua)) to luatest and adapt it.

georgiy-belyanin:
The original `cluster` module (tarantool/test/config-luatest/cluster.lua) has been moved to the current project and will be available as follows:

```lua
local t = require('luatest')
local cluster = t.cluster:new(...)

cluster:start()
```

It is used to simplify managing Tarantool clusters based on the provided configuration. The helper requires Tarantool 3.0.0 or newer. Otherwise cluster methods cause an error.

Original helper created by: alexander.turenko@tarantool.org
Test author: vdavydov@tarantool.org

Closes #368

### DIFF
--- a/luatest/cluster.lua
+++ b/luatest/cluster.lua
@@ -1,0 +1,355 @@
+--- Tarantool 3.0+ cluster management utils.
+--
+-- The helper is used to automatically collect a set of
+-- instances from the provided configuration and automatically
+-- set up servers per each configured instance.
+--
+-- @usage
+--
+-- local cluster = Cluster:new(config)
+-- cluster:start()
+-- cluster['instance-001']:exec(<...>)
+-- cluster:each(function(server)
+--     server:exec(<...>)
+-- end)
+--
+-- After setting up a cluster object the following methods could
+-- be used to interact with it:
+--
+-- * :start() Startup the cluster.
+-- * :start_instance() Startup a specific instance.
+-- * :stop() Stop the cluster.
+-- * :each() Execute a function on each instance.
+-- * :size() get an amount of instances
+-- * :drop() Drop the cluster.
+-- * :sync() Sync the configuration and collect a new set of
+--   instances
+-- * :reload() Reload the configuration.
+--
+-- The module can also be used for testing failure startup
+-- cases:
+--
+-- Cluster:startup_error(config, error_message)
+--
+-- @module luatest.cluster
+
+local fun = require('fun')
+local yaml = require('yaml')
+local assertions = require('luatest.assertions')
+local helpers = require('luatest.helpers')
+local hooks = require('luatest.hooks')
+local treegen = require('luatest.treegen')
+local justrun = require('luatest.justrun')
+local server = require('luatest.server')
+
+local Cluster = require('luatest.class').new()
+
+-- Cluster uses custom __index implementation to support
+-- getting instances from it using `cluster['i-001']`.
+--
+-- Thus, we need to change the metatable of the class
+-- with a custom __index method.
+local mt = getmetatable(Cluster)
+mt.__index = function(self, k)
+    local method = rawget(mt, k)
+    if method ~= nil then
+        return method
+    end
+
+    local server_map = rawget(self, '_server_map')
+    if server_map ~= nil and server_map[k] ~= nil then
+        return server_map[k]
+    end
+
+    return rawget(self, k)
+end
+
+local cluster = {
+    _group = {}
+}
+
+function Cluster:inherit(object)
+    setmetatable(object, self)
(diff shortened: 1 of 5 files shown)


georgiy-belyanin:
The changes I've made to the original helper:

* `g.cluster` variable preserving the cluster is now `g._cluster` to avoid clashing with the module itself in this scenario:
```lua
local t = require('luatest')
-- If the created cluster is saved in g.cluster the clash occurs
local cluster = t.cluster.new(...)
```

* Now `cluster.init`/`drop`/`clean` hooks are registered automatically (I've dropped `init()` since it only had a stub in it).
* I've changed the wait_for_start code a bit since there is a problem with capturing upvalue `luatest` by server:exec.
```diff
-             server:exec(function()
-                 t.helpers.retrying({timeout = 60}, function()
-                     t.assert_equals(box.info.status, 'running')
-                 end)
+             helpers.retrying({timeout = 60}, function()
+                 assertions.assert_equals(server:eval('return box.info.status'),
+                                          'running')
```

luatest/cluster.lua
```diff
@@ -0,0 +1,349 @@
...
+
+--- Create a new Tarantool cluster.
+--
+-- @tab g Group of tests.
+-- @tab config Cluster configuration.
+-- @tab[opt] server_opts Extra options passed to server:new().
+-- @tab[opt] opts Cluster options.
+-- @string[opt] opts.dir Specific directory for the cluster.
+-- @return table
+local function new(g, config, server_opts, opts)
```
Totktonada, line 252: 1. Should we inherit from `object` for consistency with other luatest APIs?
2. Should we change it to `cluster:new()` (same reason)?
3. I would eliminate the test group argument like it was done for `treegen`.

luatest/cluster.lua
```diff
@@ -0,0 +1,349 @@
...
+--- Ensure cluster startup error
+--
+-- Starts a all instance of a cluster from the given config and
+-- ensure that all the instances fails to start and reports the
+-- given error message.
+--
+-- @tab g Group of tests.
+-- @tab config Cluster configuration.
+-- @string exp_err Expected error message.
+local function startup_error(g, config, exp_err)
```
Totktonada, line 315: The same remark regarding eliminating the test group argument.

test/cluster_test.lua
```diff
@@ -0,0 +1,241 @@
...
+-- luatest.coverage.
+local server_opts = {
+    env = {
+        LUA_PATH = root .. '/?.lua;' ..
+            root .. '/?/init.lua;' ..
+            root .. '/.rocks/share/tarantool/?.lua',
+    }
+}
+
+local function assert_instance_running(c, instance, replicaset)
```
Totktonada, line 22: Do you consider [adjusting](https://github.com/tarantool/tarantool/blob/5ae2263852dc0630c6b531492bc6598b849f9d7c/.luacheckrc#L14-L23) `.luacheckrc`?

Totktonada: At least we can do it for a particular file.

luatest/cluster.lua
georgiy-belyanin: Applied. Thank you.

test/cluster_test.lua
georgiy-belyanin: I suggest not to touch `.luacheckrc` without specific reason. Current test code seems quite easy to read.

luatest/cluster.lua
georgiy-belyanin: Hm, I used to avoid this approach but you're right it's better to remain consistent.

1. Applied.
2. Applied.
3. Applied.

Thank you.

test/cluster_test.lua
Totktonada: OK.

Totktonada:
@georgiy-belyanin Thanks for the updates! I'm ready to merge it.

@locker Do you have any objections regarding [luatest-consistent API](https://github.com/tarantool/luatest/pull/406#discussion_r1891377127)?

Personally, I don't like all this quazi-oop, but it seems like this is the luatest way.

locker:
> @locker Do you have any objections regarding [luatest-consistent API](https://github.com/tarantool/luatest/pull/406#discussion_r1891377127)?

No.

PR #417, opened 2025-04-04
Fix deadlock in case test performs lots of checks without yielding

### LINKED ISSUE #416: luatest hangs if tests perform too many checks without yielding execution
Add the following test to `test/bad_test.lua`:

```lua
local t = require('luatest')
local g = t.group()

for i = 1, 100 do
    g['test_' .. i] = function()
        for j = 1, 100 do
            t.assert_equals(i, i)
            t.assert_equals(j, j)
        end
    end
end
```

Run it with disabled capturing (`-c`):

```
bin/luatest -c -v test/bad_test.lua
```

It hangs:

```
$ bin/luatest -c -v test/bad_test.lua
Tarantool version is 3.4.0-entrypoint-227-gf1ccb2f142f6
Started on Fri Apr  4 17:56:30 2025
luatest | 2025-04-04 17:56:30.081 [412780] main/104/luatest/luatest.log runner.lua:306 I> Bootstrap finished: 1 test(s), 0 group(s)
    bad.test_1 ...
    bad.test_1 ... (0.005s) Ok
    bad.test_69 ...
    bad.test_69 ... (0.004s) Ok
    bad.test_7 ...
```

Notes:
 - If capturing is enabled, the test passes successfully.
 - If logs are written to the unified log file (`-l <path>`), the test passes as well.
 - The issue was reproduced at commit 6c5c8b738525666ce9af0fd63dd6e7fe4fb7b0b3. At the previous commit (2f16ac767f75901d233b4cca6d9fe2ead9757609), the test passes.

locker:
Before commit 6c5c8b738525 ("Print luatest logs to stdout when unified log is disabled"), luatest implicitly configured the logger to operate in the non-blocking mode by using a pipe for the logs destination. Now, it uses a file albeit the file actually points to a pipe file descriptor. Tarantool thinks it's a regular file and writes all logs in the blocking mode. This leads to a deadlock if a test performs a lot of checks (which are logged) without yielding execution to the fiber processing logs because the size of a pipe is quite limited. Let's fix this issue by writing all logs through a pipe, as we did before. Let's also explicitly enable non-blocking logging to avoid breakages like this in future.

The fix isn't ideal because it may lead to dropping logs. We should probably handle logs emitted by the luatest runner in a separate process, like we do for processes spawned in tests.

Closes #416

### DIFF
--- a/luatest/log.lua
+++ b/luatest/log.lua
@@ -42,7 +42,12 @@ function log.initialize(options)
     output_beautifier:enable()
 
     -- Redirect all logs to the pipe created by OutputBeautifier.
-    local log_cfg = string.format('/dev/fd/%d', output_beautifier.pipes.stdout[1])
+    --
+    -- Write through a pipe to enable non-blocking mode. Required to prevent
+    -- a deadlock in case a test performs a lot of checks without yielding
+    -- execution to the fiber processing logs (gh-416).
+    local log_cfg = string.format('| cat > /dev/fd/%d',
+                                  output_beautifier.pipes.stdout[1])
 
     -- Logging cannot be initialized without configuring the `box` engine
     -- on a version less than 2.5.1 (see more details at [1]). Otherwise,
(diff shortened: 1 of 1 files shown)


Totktonada (approved):
I'm OK with it taking it as a hotfix.

I think that we can bufferize luatest's output in runtime and actually call `log.<level_func>` from a separate fiber. That's similar to the [test-run's problem](https://github.com/tarantool/test-run/pull/434) with pipe buffer size.

PR #420, opened 2025-04-08
Run tests and packaging on ubuntu-latest

locker:
Ubuntu 20.04 is no longer supported by GitHub runners.


locker:
Tarantool < 2.11 are unavailable on ubuntu-latest. Should we stop supporting those old versions in luatest?

PR #428, opened 2025-09-02
server: fix opts.reset handling in grep_log

grafin:
Before this commit `opt.reset` wasn't working, because it was rewritten with true.


luatest/server.lua
Totktonada, line 1: There is `CHANGELOG.md`, let's add an entry to it.

PR #451, opened 2026-04-01
server: set package.searchroot

### LINKED ISSUE #450: Luatest does not provide package.searchroot for created Tarantool instances
Luatest does not provide `package.searchroot` for created Tarantool instances. This leads to a situation where the instances may not have the same rocks modules as Tarantool running in the same directory where luatest is run. The most problematic is that if luatest is run using `test-run`, and rocks are not installed in the same directory (`<some_path>/.rocks/share/tarantool`) as the `luatest` module, then rocks are not properly loaded using `require()`. The reason this works for `<some_path>/.rocks/share/tarantool`, which contains `luatest`, is that `test-run` provides a LUA_PATH derived from the path to the `luatest` module. This is also the reason why `.so` modules are not properly included using `require()` (since they are located in `.rocks/lib/tarantool` and provided by `test-run` `LUA_CPATH` does not have this path).

ImeevMA:
This patch sets the `package.searchroot` value for a newly created server to the same value as in the Tarantool that created the server.

Closes #450

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+- Fixed a bug when a test Tarantool server could not find Lua modules installed
+  from rocks in the current test directory or its ascendants.
+  To avoid compatibility issues with the old behavior, the `setsearchroot`
+  server option was added (gh-450).
+
 ## 1.4.1
 
 - Fixed a bug when an error thrown by a function executed with `server:exec()`
--- a/luatest/server.lua
+++ b/luatest/server.lua
@@ -51,6 +51,8 @@ local Server = {
         alias = '?string',
 
         coverage_report = '?boolean',
+
+        setsearchroot = '?boolean',
     },
 }
 
@@ -103,6 +105,8 @@ end
 --   (but still passes `--name <alias>`).
 -- @tab[opt] object.remote_config If `config_file` is not passed, this config
 --   value is used to deduce advertise URI to connect net.box to the instance.
+-- @tab[opt] object.setsearchroot Set `package.searchroot` to be the same as
+--   in the Tarantool that starts the server.
 -- @tab[opt] extra Table with extra properties for the server object.
 -- @return table
 function Server:new(object, extra)
@@ -422,6 +426,16 @@ function Server:start(opts)
         -- instead of just `/path/to/script.lua`.
         table.insert(args, 1, command)
         command = arg[-1]
+    end
+
+    -- To fix the issue with rock-modules missing from the created instances,
+    -- we need to set `package.searchroot`. Furthermore, this must be done
+    -- before loading `command`. This is done if the `setsearchroot` option
+    -- is true or missing.
+    if self.setsearchroot == nil or self.setsearchroot then
+        table.insert(args, 1, '-e')
+        table.insert(args, 2,
+            string.format("package.setsearchroot('%s')", package.searchroot()))
     end
 
     local log_cmd = {}
--- a/test/server_test.lua
+++ b/test/server_test.lua
@@ -20,9 +20,6 @@ local server = Server:new({
     command = command,
     workdir = fio.pathjoin(datadir, 'common'),
     env = {
-        LUA_PATH = root .. '/?.lua;' ..
-            root .. '/?/init.lua;' ..
-            root .. '/.rocks/share/tarantool/?.lua',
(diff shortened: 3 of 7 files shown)


luatest/server.lua
```diff
@@ -611,6 +611,9 @@ function Server:wait_until_ready()
         end)
         return ok and is_ready
     end)
+
+    self.net_box:eval(string.format("package.setsearchroot('%s')",
+        package.searchroot()))
```
locker: I don't like that this action is performed in `wait_until_ready`. Can we pass it in the `-e` command line argument?

Also, please
- add a comment explaining why this is done;
- make sure that all Tarantool tests pass after this change.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Fixed a bug where the server could not find the same rock modules
+  as the Tarantool that created the server (gh-450).
```
locker: Let's rephrase: "Fixed a bug when a test Tarantool server could not find Lua modules installed from rocks in the current test directory or its ascendants (gh-450)."

luatest/server.lua
ImeevMA: Thanks, done.

CHANGELOG.md
ImeevMA: Thanks, fixed. Also, since we decided to add a new option for `server.start()`, I've added its description here.

luatest/server.lua
```diff
@@ -401,8 +401,13 @@ end
 -- @bool[opt] opts.wait_until_ready Wait until the server is ready.
 --   Defaults to `true` unless a custom executable was provided while building
 --   the server object.
+-- @bool[opt] opts.set_search_root Set `package.searchroot` to be the same as
+--   in the Tarantool that starts the server.
 function Server:start(opts)
-    checks('table', {wait_until_ready = '?boolean'})
+    checks('table', {
+        wait_until_ready = '?boolean',
+        set_search_root = '?boolean',
```
locker: This option logically belongs in `Server:new()`, not in `Server:start()`. Also, better remove underscores from its name to be consistent with the `package.setsearchroot()` function name.

luatest/server.lua
```diff
@@ -424,6 +429,16 @@ function Server:start(opts)
...
     end
 
+    -- To fix the issue with rock-modules missing from the created instances,
+    -- we need to set `package.searchroot`. Furthermore, this must be done
+    -- before loading `command`. This is done if the `set_search_root` option
+    -- is true or missing.
+    if opts == nil or opts.set_search_root == nil or opts.set_search_root then
+        table.insert(args, 1, '-e')
+        table.insert(args, 2,
+            string.format("package.setsearchroot('%s')", package.searchroot()))
```
locker, line 438: Is it really necessary to pass the indices to `table.insert()`?

test/server_test.lua
```diff
@@ -753,3 +753,18 @@ g.test_error_in_app_thread = function()
...
+g.before_test('test_pass_package_searchroot_to_server', function()
+    g.s = Server:new()
+    g.s:start()
+end)
+
+g.after_test('test_pass_package_searchroot_to_server', function()
+    g.s:drop()
+end)
+
+g.test_pass_package_searchroot_to_server = function()
```
locker: Instead of adding a new test, wouldn't be better to simply remove `LUA_PATH` from the test server created at the global level?

luatest/server.lua
ImeevMA: Thanks, fixed.

luatest/server.lua
ImeevMA: Yes, we want `-e 'package.setsearchroot()'` to be executed before the `command`.

test/server_test.lua
ImeevMA: True, thanks. Fixed.

test/server_test.lua
```diff
@@ -20,9 +20,6 @@ local server = Server:new({
     command = command,
     workdir = fio.pathjoin(datadir, 'common'),
     env = {
-        LUA_PATH = root .. '/?.lua;' ..
```
locker, line 23: Let's remove `LUA_PATH` from all tests.

ImeevMA: Thanks, fixed. Also, I removed my test.

locker (approved):

PR #454, opened 2026-04-13
Forbid using with Tarantool < 2.2.1

### LINKED ISSUE #453: Luatest 1.4.2 has broken the compatibility with Tarantool 1.10
Tarantool 1.10 cannot be used with Luatest anymore due to https://github.com/tarantool/luatest/commit/bffdd7d7210093446b3d93b18a8e694c8e8fa17a. Any test fails with error:

```
not ok 1	util.test_uri_eq
#   .../vshard/vshard/.rocks/share/tarantool/luatest/server.lua:438: attempt to call field 'searchroot' (a nil value)
#   stack traceback:
#   	./test/luatest_helpers/server.lua:154: in function 'start'
#   	./test/unit-luatest/util_test.lua:12: in function <./test/unit-luatest/util_test.lua:10>
```

Tarantool 1.10 doesn't have `package.searchroot` defined, let's won't break backward compatibility, when it costs us nothing.

Serpentian:
Tarantool 1.10 cannot be used with Luatest anymore due to tarantoool/luatest@bffdd7d. Any test fails, since Tarantool 1.10 doesn't have `package.searchroot` defined.

Let's won't break backward compatibility, when it costs us nothing. For that we set the `searchroot` argument iff `package.searchroot` is defined.

Closes #453

### DIFF
--- a/luatest/init.lua
+++ b/luatest/init.lua
@@ -2,6 +2,11 @@
 -- @module luatest
 local luatest = setmetatable({}, {__index = require('luatest.assertions')})
 
+local utils = require('luatest.utils')
+if not utils.version_current_ge_than(2, 2, 1) then
+    error('Luatest supports only Tarantool >= 2.2.1')
+end
+
 luatest.Process = require('luatest.process')
 luatest.VERSION = require('luatest.VERSION')
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Explicitly forbid using Luatest with Tarantool < 2.2.1 (gh-453).
+
 ## 1.4.2
 
 - Fixed a bug when a test Tarantool server could not find Lua modules installed


ImeevMA (approved):
Hi! Thank you for the patch! I have just one minor comment - I think there should `if` instead of `iff` in commit-message and title.

Serpentian:
`iff` means `if and only if`, it's pretty widely used)

luatest/server.lua
```diff
@@ -432,7 +432,8 @@ function Server:start(opts)
     -- we need to set `package.searchroot`. Furthermore, this must be done
     -- before loading `command`. This is done if the `setsearchroot` option
     -- is true or missing.
-    if self.setsearchroot == nil or self.setsearchroot then
+    if package.searchroot and
+       (self.setsearchroot == nil or self.setsearchroot) then
```
locker: I really don't like that this makes luatest behavior depend on Tarantool version. See, we're planning to remove the [code](https://github.com/tarantool/test-run/blob/c77851d09266e8923a1c6d7fdff2d170178c78ea/bin/luatest) setting `LUA_PATH` from `test-run` because it seems that it was added because of the bug in luatest fixed with `package.setsearchroot`. If we do that, you won't be able to bump test-run in vshard.

Tarantool 1.10 is more that two years past end-of-support. Do you really need to test vshard against it?

Serpentian: Nah, if we plan to remove that code, not much sense in that PR. We can drop support for 1.10 anytime in vshard.

locker: You can also pin to an older luatest version, like we do in Tarantool, but you won't be able to adopt new luatest features then.

Serpentian: Let's then explicitly forbid using Luatest and not throw some opaque error)

luatest/init.lua
```diff
@@ -1,3 +1,8 @@
+local utils = require('luatest.utils')
```
locker, line 5: This change breaks the generate-docs workflow.

Serpentian: Fixed

locker (approved):

PR #463, opened 2026-05-05
pp: implement verbose serialization of box errors

locker:
When a test fails with an error, the error is serialized with the `tostring` Lua function, which by default prints only the error message. This means that some useful information, such as the error trace and payload fields, are lost. We could enable verbose error serialization using the `box_error_serialize_verbose` compat module option, but it may break existing tests. Let's instead unpack errors thrown by tests before printing them to make debugging easier.

Closes #445

### DIFF
--- a/luatest/pp.lua
+++ b/luatest/pp.lua
@@ -1,5 +1,6 @@
 local Class = require('luatest.class')
 local sorted_pairs = require('luatest.sorted_pairs')
+local utils = require('luatest.utils')
 
 -- Pretty printer.
 local pp = {
@@ -160,6 +161,8 @@ function Formatter.mt:format(v, indentLevel)
                 return tostring(i)
             end
         end
+    elseif utils.is_box_error(v) then
+        return self:format_table(utils.error_unpack(v), indentLevel)
     end
 
     return tostring(v)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Implemented verbose serialization of uncaught box errors thrown in tests
+  (gh-445).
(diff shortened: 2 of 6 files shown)


luatest/pp.lua
```diff
@@ -138,6 +139,9 @@ function Formatter.mt:format_table(tbl, indentLevel)
 end
 
 function Formatter.mt:format(v, indentLevel)
+    if utils.is_box_error(v) then
+        v = utils.error_unpack(v)
+    end
```
nshy: Nit. Looks like adding another `elseif` branch will be more straightforward. Up to you)

nshy (approved):

luatest/pp.lua
locker: Makes sense. Done, thanks!

<details><summary>diff</summary>

```diff
diff --git a/luatest/pp.lua b/luatest/pp.lua
index a2b1629d6bf6..ab5b181f1a6b 100644
--- a/luatest/pp.lua
+++ b/luatest/pp.lua
@@ -139,9 +139,6 @@ function Formatter.mt:format_table(tbl, indentLevel)
 end
 
 function Formatter.mt:format(v, indentLevel)
-    if utils.is_box_error(v) then
-        v = utils.error_unpack(v)
-    end
     local type_v = type(v)
     if "string" == type_v  then
         return string.format("%q", v)
@@ -164,6 +161,8 @@ function Formatter.mt:format(v, indentLevel)
                 return tostring(i)
             end
         end
+    elseif utils.is_box_error(v) then
+        return self:format_table(utils.error_unpack(v), indentLevel)
     end
 
     return tostring(v)
```

</details>

PR #468, opened 2026-06-25
Disable system checks by default in `cbuilder` base_config

Godrik0:
This PR sets `config.checks` to `'off'` in `cbuilder.base_config` to avoid environment-dependent alerts (THP, readahead) in tests that use `cbuilder`.

Tests that need checks enabled should override `config.checks` explicitly.

Needed for tarantool/tarantool#12371.

### DIFF
--- a/luatest/cbuilder.lua
+++ b/luatest/cbuilder.lua
@@ -97,6 +97,16 @@ local base_config = {
         -- by half for my test.
         timeout = 0.1,
     },
+    -- Disable system checks by default in tests to avoid
+    -- environment-dependent alerts (THP, readahead, etc.).
+    conditional = {
+        {
+            ['if'] = 'tarantool_version >= 3.8.0',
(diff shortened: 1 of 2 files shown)


luatest/cbuilder.lua
```diff
@@ -97,6 +97,16 @@ local base_config = {
...
+    -- Disable system checks by default in tests to avoid
+    -- environment-dependent alerts (THP, readahead, etc.).
+    conditional = {
+        {
+            ['if'] = 'tarantool_version >= 3.8.0',
+            config = {
+                checks = 'off',
+            },
+        },
+    },
```
Totktonada, line 109: It possibly worth to implement a more granular check instead of a version-based one: https://github.com/tarantool/metrics/pull/553. This way we handle tarantool developer builds (non-released intermediate commits) as well.

PR #473, opened 2026-07-03
treegen: create temporary files in VARDIR

### LINKED ISSUE #467: Treegen helper pollutes system tmp directory
The `treegen` helper creates test directories right in `/tmp`:

https://github.com/tarantool/luatest/blob/09a627697564a72f6ac45f5ca632f6580e30bdd6/luatest/treegen.lua#L141-L162

This is annoying because if the test crashes or the `KEEP_DATA` environment variable is set, the test directories will remain in `/tmp` and removing them manually is tedious because they use generic names.

Let's create temporary directories in the `VARDIR` (`/tmp/t` by default), like the `server` helper does. Also, it'd be nice to give them some readable names, for example, `justrun-XXXX`.

Related issue: #411.

locker:
The `treegen` module uses `fio.tempdir()` for creating temporary directories. The latter uses `/tmp` as the base directory. This is bad because if a test fails or the KEEP_DATA environment variable is set, temporary directories are not removed, thus polluting `/tmp`.

Let's override TMPDIR to VARDIR to force `fio.tempdir()` use VARDIR (`/tmp/t` by default) for creating temporary files.

Closes #467

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   were ignored (gh-465).
 - Fixed a bug when the luatest log file was removed if the command line
   option `--no-cleanup` was not used (gh-471).
+- Now the `treegen` helper creates temporary files in `VARDIR`, which is
+  `/tmp/t` by default, so as not to pollute the system tmp directory.
 
 ## 1.4.4
 
--- a/luatest/runner.lua
+++ b/luatest/runner.lua
@@ -306,6 +306,10 @@ function Runner.mt:run()
     end
     fio.mktree(Server.vardir)
 
+    -- Set TMPDIR to make fio.tempdir() create temporary files in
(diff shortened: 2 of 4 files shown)


Totktonada:
We may want to create directories outside of `/tmp/t` if `KEEPDIR` is enabled, because `/tmp/t` will be pruned by the next test run.

nshy (approved):

locker:
> We may want to create directories outside of `/tmp/t` if `KEEPDIR` is enabled, because `/tmp/t` will be pruned by the next test run.

You can always use the `--no-clean` luatest option to avoid that. Or you can just copy `/tmp/t` to a safe location if you want to examine artifacts. Note that test-run uses a unique `VARDIR` for each test worker.

